PR #71, opened 2022-07-02
Fix support string expr

k2tzumi:
# Problem

The value to be expanded into an expression is determined to be a number, and single quotes are added to the path and GET parameters.

* Example
  ```yaml
  vars:
    year: 2022
  steps:
    -
      req:
        /api/path?year={{ vars.year }}:
          get:
  ```

* Execution logs
  ```
  -----START HTTP REQUEST-----
  GET /api/path?year=%272022%27 HTTP/1.1
  ```

## Proposal for revision

If the expanded content after executing string() is not only that expression, it seems better not to stringify it.

## Related PR

https://github.com/k1LoW/runn/pull/21

### DIFF
--- a/operator.go
+++ b/operator.go
@@ -697,7 +697,8 @@ func (o *operator) expand(in interface{}) (interface{}, error) {
 			var s string
 			switch v := o.(type) {
 			case string:
-				if numberRe.MatchString(v) {
+				// Stringify only one expression.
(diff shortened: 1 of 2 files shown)


k1LoW:
@k2tzumi  You're absolutely right! Thank you!

k1LoW:
Released as v0.22.3.

PR #73, opened 2022-07-03
Support OpenAPI3 validation at authentication request

k2tzumi:
## Reasons for Proposal

I want to validate HTTP requests with authentication.

Verification currently yields the following error.

```log
openapi3 validation error: Security requirements failed
```

## Proposal for revision

Since checking the response code is sufficient to verify the authentication itself, you may want to set an option not to perform authentication checks when verifying the request.

NoopAuthenticationFunc was used.

see.
https://github.com/getkin/kin-openapi/blob/648d6b9a170b6162c79927aeba39bda2fe37386a/openapi3filter/validation_handler.go#L17-L18

### DIFF
--- a/http_validator.go
+++ b/http_validator.go
@@ -150,6 +150,9 @@ func (v *openApi3Validator) requestInput(req *http.Request) (*openapi3filter.Req
 		Request:    req,
 		PathParams: pathParams,
 		Route:      route,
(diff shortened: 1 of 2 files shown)


k1LoW (approved):
@k2tzumi Thank you!! Looks Great!!!

PR #81, opened 2022-07-10
Support nil & bool expand

k2tzumi:
# Problem

If the value to be expanded into an expression is bool or nil, an error will occur.

* Error example
  * boolean
    ```log
    invalid format: true
    ```
  * nil
    ```log
    invalid format: <nil>
    ```

# Proposal for revision

It would be good to expand appropriately when the value is bool or nil.

### DIFF
--- a/operator.go
+++ b/operator.go
@@ -771,6 +771,10 @@ func (o *operator) expand(in interface{}) (interface{}, error) {
 				s = strconv.Itoa(int(v))
 			case int:
 				s = strconv.Itoa(v)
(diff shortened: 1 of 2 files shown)


k2tzumi:
Oops.
When I did the nil expansion, it was indistinguishable from trying to expand a variable that does not exist.
I am wondering if separate error handling is needed.

k2tzumi:
PR resubmitted.
https://github.com/k1LoW/runn/pull/82

PR #82, opened 2022-07-10
Support bool expand

k2tzumi:
# Problem

If the value to be expanded into an expression is bool, an error will occur.

* Error example
  ```log
  invalid format: true
  ```

# Proposal for revision

It would be good to expand appropriately when the value is bool.


k1LoW:
@k2tzumi Thank you!

PR #89, opened 2022-07-23
Support bash like interpolate 

k2tzumi:
# Proposal

Allow default values to be specified when expanding environment variables in yaml.

* Example
```yml
desc: Example
debug: ${DEBUG:-true}
interval: ${INTERVAL:-5}
```

see.
https://github.com/k1LoW/expand/pull/4

# Contents of Proposal

* Bump up packages.
* Append test.

### DIFF
--- a/book_test.go
+++ b/book_test.go
@@ -1,6 +1,8 @@
 package runn
 
 import (
+	"os"
+	"strconv"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


k1LoW:
@k2tzumi Thank you!!!

PR #90, opened 2022-07-30
csvq added to support databases

k2tzumi:
# Proposal

Add [csvq](https://github.com/mithrandie/csvq) to the support database to handle the following formats for test data

* Support following file formats
  * CSV
  * TSV
  * LTSV
  * JSON
  * etc..

* Example
  * book
    ```yml
    desc: Test using csvq
    runners:
      db: csvq://./testdata/csvq
    steps:
      -
        db:
          query: SELECT * FROM csv_data WHERE value > 0
      -
        test: steps[0].rows[0].id == "2"
    ```
  * testdata/csvq/csv_data.csv
    ```csv
    id,name,value
    1,a,0
    2,b,1
    3,c,1
    ```

see.
https://github.com/k1LoW/runn/commit/344b07ce92c915e52164d860c2548dfd5e46823a

# Contents of Proposal

* [import csvq-driver.](https://github.com/k1LoW/runn/commit/cb8b2e0b58efb79c2cad927b690b436c4e49c666)
* Append test.


k1LoW:
@k2tzumi 

It is an interesting feature.

I have one question.
Can you share a use case for using csvq?
I could not think of any.

Depending on the use case, a different implementation might be better.

k1LoW:
Possibly an improvement in `vars:` would be more useful for data-driven testing than an improvement in DB Runner?

Even now it is possible to load JSON data as follows

``` go
func TestUsingJSON(t *testing.T) {
	ctx := context.Background()
	b := []byte(`{"name":"hoge","age":30}`)
	var d interface{}
	if err := json.Unmarshal(b, &d); err != nil {
		t.Fatal(err)
	}
	opts := []runn.Option{
		runn.T(t),
		runn.Book("path/to/runbook.yml"),
		runn.Vars("json_data", d),
	}
	o, err := runn.New(opts...)
	if err != nil {
		t.Fatal(err)
	}
	if err := o.Run(ctx); err != nil {
		t.Fatal(err)
	}
}

**Comment by k1LoW** [comment]
> A single endpoint can have many combinations of parameters and their expected values.

I think we also need a syntax for loops, like getting values from the data one by one and using them.

**Comment by k2tzumi** [comment]
The idea of improving vars: looks very good.
It would be easier to do data-driven enhancements if we could handle JSON objects instead of JSON strings.

To make it easier to handle large JSON objects, the following extensions could be considered

* Ability to read data outside of yaml
* Ability to add and rewrite portions of JSON objects

Either or both of the above would make it easier to manage data.

Loop syntax may also be needed somewhere.
I think the ease of repeated testing would be improved if it were possible to use Include to expand multiple patterns and replace the request data and expected values while rewriting the var.

**Comment by k2tzumi** [comment]
> Even now it is possible to load JSON data as follows

How about defining built-in functions to manipulate JSON strings and objects?
For example, a function like the following

* JsonUnmarshal
convert a JSON string into a JSON object and use it in the request body
* JsonReplace, JsonMerge...
Used to rewrite JSON objects

Instead of defining a custom function in go test, it would be easier to test in yaml alone by defining a standard function in advance.
The policy is to extend existing functions by combining built-in functions.
This would require only the addition of a function to map JSON objects to request bodies in HTTPRunner.
However, this will increase runn’s responsibility, and we believe this will affect the policy for future function extensions, so we think careful judgment is necessary.

**Comment by k1LoW** [comment]
> Ability to read data outside of yaml

I thought of the following ideas

The `vars:` section supports schemes to read external files such as `json://` or `file://`.

```yaml
vars:
   test_data: json://path/to/test.json
```

If it can be loaded into the vars section, it can then be used as an object.

https://go.dev/play/p/qgqk5Ay6ITF

This can be done the same way for YAML and for LTSV.

( I'm not sure whether to use `json://`, `yaml://`, and `csv://`, or whether to use `file://` to determine automatically. )

The bulit-in function can also be used universally by preparing one for an object.

---

> This would require only the addition of a function to map JSON objects to request bodies in HTTPRunner.

This is a very interesting idea!!!

k2tzumi:
I thought that if a specific key in the vars section is used to read a file, there would be restrictions on the names of variables that can be used.

I think it would be better to define `include-vars` as a new section.
This would allow us to use the same variable names as before, define objects when reading, and add options for partial rewriting.
Also, it looks like it will be possible to load multiple files.

* When loading multiple files
  ```yml
  vars:
     include_path: "json://path/to/include"
  include_vars:
    - 
      name: "request_vars"
      path: "{{ $vars.include_path }}/request_body.json"
    -
      name: "expected_vars"
      path: "{{ $vars.include_path }}/expected.json" 
  ```
   Can be referenced by `vars.request_vars` and `vars.expected_vars` respectively

* When rewriting a part of json
  ```yml
  vars:
     replace_var: "replaced"
  include_vars:
    - 
      name: "include_vars"
      path: "json://path/to/include/test.json"
      replace: ".foo|='{{ var.replace_var }}'"  # For example, jq like replace
  ```
 

I think it would be easier to reuse runbooks by making effective use of Include Runner when used with existing vars.

k1LoW:
> I thought that if a specific key in the vars section is used to read a file, there would be restrictions on the names of variables that can be used.

Indeed, it will no longer accept `json://xxx` as a string, but this is a rare case.

Accepting multiple files can also be expressed as follows

```yaml
vars:
  request_vars: "json://path/to/request_body.json"
  expected_vars: "json://path/to/expected.json"
steps:
  get_body:
    dump: request_vars[0].req.body
  get_expected:
    dump: expected_vars[0].expected
```

I think it is better to use special handling only if you want to accept `json://xxx` as a string.

```yaml
vars:
  request_vars: 
    value: "json://path/to/request_body.json"
    as_value: true # Setting name `as_value` is appropriate.
steps:
  get_value:
    dump: request_vars # => 'json://path/to/request_body.json'
```

I also think that REPLACEMENT and FILTERING are not the responsibility of the `vars:` section. 

It would be better to add built-in functions that do replacement and filtering, which can be used for other values as well.

k2tzumi:
> Accepting multiple files can also be expressed as follows

I was under the mistaken impression that the variable names were fixed and that there was a limit to the number of files that could be read.
The sample code helped me understand. Thank you very much.

> I think it is better to use special handling only if you want to accept json://xxx as a string.

It's a rare case, so it doesn't seem to need that special treatment.

> I also think that REPLACEMENT and FILTERING are not the responsibility of the vars: section.
>
> It would be better to add built-in functions that do replacement and filtering, which can be used for other values as well.

I thought that was correct.
Since the vars section is designed to be simple with no expression expansion, it would be nice to have built-in functions available in other sections.

k2tzumi:
@k1LoW 

Thank you for always taking in PRs so quickly.

We feel that the advice we have received has brought us closer to achieving data-driven testing.

With the current functionality, the following configuration is close to the image.

* main scenario
  ```yml
  desc: main scenario
  vars:
    requestBody:
      default: "1"
    expect: null
  steps:
    -
      req:
        /post:
          post:
            body:
              application/json: "{{ vars.requestBody }}"
    -
      test: |
        steps[0].res.status == 200
        && steps[0].res.body== vars.expect
  ```
* loop scenario
  ```yml
  desc: loop scenario
  steps:
    -
    include:
      path: path/to/main_scenario.yml
      vars:
        requestBody: json://path/to/case1.json
        expect: json://path/to/expect1.json
    -
    include:
      path: path/to/main_scenario.yml
      vars:
        requestBody: json://path/to/case2.json
        expect: json://path/to/expect2.json
  
  .. snip ..
  ```

> I think we also need a syntax for loops, like getting values from the data one by one and using them.

The next step would be to want a loop syntax.
In the above example, the amount of description was reduced by combining json file loading and includes, but it could be improved a little more.

I have been thinking about which direction to go with the functionality expansion, and would like your input on which would be more suitable.

1. Extend Include Runner
2. Extend retry settings for steps
3. Add a new loop syntax

Which proposal seems to be better?

k1LoW:
I would like to select "Add a new loop syntax".

The features I would like to see in the new syntax are

1. Execute step a specified number of times
2. Can get and use indexes during iteration

### 1. Execute step a specified number of times

I would like to be able to iterate on every step, not just `include runner`.

I believe the syntax position would be the same as the `retry:`.

### 2. Can get and use indexes during iteration

Within a step, I would like to be able to receive the index of iterations as a variable (at least. I would like to get the "number" of iterations).

## Syntax image

(This is just a syntax image that I am currently thinking of, and still needs to be considered.)

```yaml
steps:
  echo:
    loop: 10
    exec: echo {{ i }}
```

```yaml
vars:
  requestBodies: json://path/to/cases.json
  expects: json://path/to/expects.json
steps:
  -
    include:
      loop: len(requestBodies)
      path: path/to/main_scenario.yml
      vars:
        requestBody: requestBodies[i]
        expect: expects[i]
```

k2tzumi:
> I believe the syntax position would be the same as the retry:.

I understood the direction to support the following syntax

*`steps[*].loop:`
*`steps.<key>.loop:`

> I would like to be able to iterate on every step, not just include runner.

I thought it was a very good idea for a permanent solution.

The only concern I have is that the retry syntax and use cases would be the same, which would increase complexity.

For example, if loop and retry are specified at the same time, which is evaluated first? etc.

> Can get and use indexes during iteration

I thought it would be nice if the above extension could be done with the retry syntax.

I could not think of a reason to make a clear distinction between the use cases for loop syntax and retry syntax.

We may want to review the name of the retry syntax to clarify the intent.

k2tzumi:
@k1LoW 

I have a question.

> It would be better to add built-in functions that do replacement and filtering, which can be used for other values as well.

When implementing the built-in functionality, do we need the ability to disable the built-in functionality as a runtime option for the runn command?

```console
$ runn run --disable-built-in-fuctions path/to/scenario.yml
```

How to deal with the impact of adding custom functions from go test? I'd like to think about it.

https://github.com/k1LoW/runn/tree/main#example-add-custom-function--func-func-

* Would you like to enable built-in functions only at CLI runtime?
* Only call custom functions in preference to built-in functions
No additional controls are provided.
* Explicitly implement an option to disable built-in functions and have it selected when calling from the code base
book's structure has a built-inFunctions flag

I am thinking the last idea looks good and easy to understand

k1LoW:
> We may want to review the name of the retry syntax to clarify the intent.

I think `retry:` is a syntax to retry when a step fails.

`loop:` is a syntax for running step iteratively without conditions.

I would like to separate these 2 syntaxes if possible (I believe these are created in a separate feature in many codes as well).

> For example, if loop and retry are specified at the same time, which is evaluated first? etc.

Yes, there is indeed a problem as you say.

The ideas I have are as follows

1. Think of a nice name that has both `retry:` and `loop:` features.
2. Exclusive control of `retry:` and `loop:` in the same step.
    - Using "Include runner", both `retry in loop` and `loop in retry` can be implemented.
    - For now, I prefer to select this one
3. Determine the order of execution by statement position.
    - They are the same indentation position, so I don't really want to select...
4. Anything else...

k1LoW:
@k2tzumi 

https://github.com/k1LoW/runn/pull/90#issuecomment-1214874427

If we were to make it behave like a `vars:` section, that would be the second option.

> Only call custom functions in preference to built-in functions
> No additional controls are provided.

There are times when we do not want to disable all built-in functions. Although it is a little difficult to understand, how about allowing functions to be overwritten?

k1LoW:
https://github.com/k1LoW/runn/pull/90#issuecomment-1214876144

I'm starting to think that loop and retry can be combined...

> 1. Think of a nice name that has both retry: and loop: features.

k2tzumi:
https://github.com/k1LoW/runn/pull/90#issuecomment-1214876144

> I think retry: is a syntax to retry when a step fails.

I thought that `HTTP Runner` and `gRPC Runner` would be the only use cases where steps fail due to external factors and are recovered by retry.

However, I think it is a little confusing that all steps are retryable at present.

> I would like to separate these 2 syntaxes if possible (I believe these are created in a separate feature in many codes as well).

If the syntax is separated, I thought it would be preferable to limit the specification of the retry syntax to only `HTTP Runner` and `gRPC Runner`.

However, considering the cost of separating the implementations, it is a distressing issue.

> I'm starting to think that loop: and retry: can be combined...

I'm thinking the same way myself.

k2tzumi:
https://github.com/k1LoW/runn/pull/90#issuecomment-1214947315

> If we were to make it behave like a vars: section, that would be the second option.

Let's go with a simple implementation.

> There are times when we do not want to disable all built-in functions. Although it is a little difficult to understand, how about allowing functions to be overwritten?

Yes. We see no problem if we can override the function

k1LoW:
@k2tzumi 
Now that we have implemented the `json://` and `loop:` sections, I'm going to close this PR.

Thank you for your proposal!!

k2tzumi:
It could be used to read the source for data-driven tests, but I have no plans to use it at this time, so I'm closing it for now.

PR #91, opened 2022-07-31
Improvement of unstable operation during OAS Spec verification

k2tzumi:
# Problem

When performing OAS Spec validation, the following error would occur in rare cases when submitting a request.

* example
```log
http request failed on ‘api test’.steps.oas: Put “http://localhost:8000/api/hoge/394336045384073216?date=202104”: http: ContentLength=591 with Body length 616
```

I couldn't reproduce it in the test code, but it looked like the request body object was being used all over again during validation.

Incidentally, the test scenario in which the problem occurred was reproducible, as it made extensive use of includes.

# Contents of Proposal

It seems to be stable if request doby is reset after verifying OAS Spec but before sending actual request.

Fixed an error in the response definition of the OAS Spec (Github API) that was found when creating the test code.

### DIFF
--- a/http.go
+++ b/http.go
@@ -149,6 +149,17 @@ func (rnr *httpRunner) Run(ctx context.Context, r *httpRequest) error {
 		if err := rnr.validator.ValidateRequest(ctx, req); err != nil {
 			return err
 		}
+		// reset Request.Body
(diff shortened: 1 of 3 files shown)


k1LoW:
@k2tzumi Thank you!!!

PR #93, opened 2022-08-11
Support for read json in var section

k2tzumi:
refs https://github.com/k1LoW/runn/pull/90#issuecomment-1207325344

# Proposal

Extend the vars section to enable data-driven testing, so that the test case yaml can be separated from the test data portion.

* Example
  ```yml
  vars:
     test_data: json://path/to/test.json
  ```

see.
https://github.com/k1LoW/runn/commit/4acccf0546becf939ddcb03106cbab0eff4126d7

# Contents of Proposal

added support for reading JSON files and referencing them as objects when the first-level key value in the vars section starts with `json://~`.


k1LoW (approved):
@k2tzumi Thank you for your GREAT commit!!!

PR #95, opened 2022-08-13
Convert number var to json compatible type

k2tzumi:
# Problem

There was a problem in which the type of the numerical value in the vars defined in yaml was different from the type of the numerical item in the response, causing the test to fail.

The reason is that the libraries used have different behavior when unmarshaling json and yaml.

* Verification Code
    ```go
    package main
    
    import (
	    "fmt"
    
	    "github.com/goccy/go-json"
	    "github.com/goccy/go-yaml"
	    "github.com/google/go-cmp/cmp"
    )
    
    func main() {
	    jb := []byte(`{"one": {"name": "alice", "age":30 }}`)
	    var ji map[string]interface{}
	    if err := json.Unmarshal(jb, &ji); err != nil {
		    panic(err)
	    }
	    yb := []byte(`
    one:
      name: "alice"
      age: 30
    `)
	    var yi map[string]interface{}
	    if err := yaml.Unmarshal(yb, &yi); err != nil {
		    panic(err)
	    }
	    if diff := cmp.Diff(ji, yi, nil); diff != "" {
		    fmt.Printf("%v", diff)
	    }
    }
    ```

* Execution logs
  ```log
    map[string]any{
  	  "one": map[string]any{
  - 		"age":  float64(30),
  + 		"age":  uint64(30),
  		  "name": string("alice"),
  	  },
    }
  
  Program exited.
  ```

I thought we needed to match the behavior when loading yaml, but am I being overly concerned?

The following scenarios test failed
https://github.com/k1LoW/runn/blob/7a826fc128a7e89c9b3435a712659088cbd73fc8/testdata/book/httpbin.yml
https://github.com/k2tzumi/runn/runs/7816805003?check_suite_focus=true#step:6:63

## Proposal for revision

* [Only vars values are reconverted to reproduce json behavior.](https://github.com/k1LoW/runn/commit/54b78d524b38352b393346dfcbdcd2080e2c6d43)
* [Numeric items are now float type, so the expansion process has been changed.](https://github.com/k1LoW/runn/commit/36f73146afaef99dfc8580fc571bce135d3cf33d)
* [Unified the library used for json processing to goccy/go-json to match the behavior.](https://github.com/k1LoW/runn/commit/cf7b6a8b42846678d2fbc67d9d4f70f05c71833e) 
Currently, `goccy/go-json` is compatible with `encoding/json`, but since most of the processing is done by `goccy/go-json`, we thought it would be better to unify them.
* [Refine invalid format message.](https://github.com/k1LoW/runn/commit/fc27e57ef6e9dcf98b37bae248583ebc50eed247)
* [httpbin](https://github.com/postmanlabs/httpbin) can now be integrated into CI for integration testing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,9 @@ jobs:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     services:
       httpbin:
-        image: kennethreitz/httpbin
+        image: kennethreitz/httpbin:latest
+        ports:
+          - 8080:80
     steps:
       - name: Set up Go ${{ matrix.go_version }}
         uses: actions/setup-go@v3
@@ -35,7 +37,7 @@ jobs:
       - name: Run tests
         run: make ci
         env:
-          HTTPBIN_END_POINT: http://httpbin/
+          HTTPBIN_END_POINT: http://localhost:8080/
 
       - name: Run octocov
         uses: k1LoW/octocov-action@v0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,9 @@ jobs:
         go_version: [1.18]
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    services:
+      httpbin:
+        image: kennethreitz/httpbin
     steps:
       - name: Set up Go ${{ matrix.go_version }}
         uses: actions/setup-go@v3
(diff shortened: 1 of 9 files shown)


k2tzumi:
On a different note, TestShard is now Flaky Test.

https://github.com/k1LoW/runn/runs/7818107956?check_suite_focus=true#step:7:56
https://github.com/k2tzumi/runn/runs/7817815730?check_suite_focus=true#step:7:58

k1LoW (approved):
LGTM!! Thank you!!

k1LoW:
> On a different note, TestShard is now Flaky Test.

Thank you. I confirm this issue too... 😕

PR #97, opened 2022-08-17
[BREAKING] Create a new `loop:` section and integrate it with the features of the `retry:` section.

k1LoW:
From https://github.com/k1LoW/runn/pull/90#issuecomment-1216396709, I would like to create a new `loop:` section and integrate it with the features of the `retry:` section.

---

### `steps[*].loop:` `steps.<key>.loop:`

Loop settings for steps.

#### Simple loop usage

``` yaml
steps:
  multicartin:
    loop: 10
    req:
      /cart/in:
        post:
          body:
            application/json:
              product_id: "{{ i }}" # The loop count (0..9) is assigned to `i`.
[...]
```

or

``` yaml
steps:
  multicartin:
    loop:
      count: 10
    req:
      /cart/in:
        post:
          body:
            application/json:
              product_id: "{{ i }}" # The loop count (0..9) is assigned to `i`.
[...]
```

#### Retry

It can be used as a retry mechanism by setting a condition in the `until:` section.

If the condition of `until:` is met, the loop is broken without waiting for the number of `count:` to be run.

Also, if the run of the number of `count:` completes but does not satisfy the condition of `until:`, then the step is considered to be failed.

``` yaml
steps:
  waitingroom:
    loop:
      count: 10
      until: 'steps.waitingroom.res.status == "201"' # Store values of latest loop
      minInterval: 0.5 # sec
      maxInterval: 10  # sec
      # jitter: 0.0
      # interval: 5
      # multiplier: 1.5
    req:
      /cart/in:
        post:
          body:
[...]
```

( `steps[*].retry:` `steps.<key>.retry:` are deprecated )


README.md
```diff
@@ -355,16 +355,55 @@ steps:
 [...]
 ```
 
-### `steps[*].retry:` `steps.<key>.retry:`
+### `steps[*].loop:` `steps.<key>.loop:`
```
k1LoW, line 358: @k2tzumi 
To share the implementation image, I updated the README first.

I would like to hear your opinion.

k2tzumi: I felt strong feelings 👍 

We believe that even though it will be difficult to integrate with the functionality of the retry: section, we will try to make sure that it is not on the same level of hierarchy as the loop: section.

```suggestion
### `steps[*].[req|greq].retry:` `steps.<key>.[req|greq].retry:`
```

We believe it will be like the above.

k1LoW: I use `retry:` for both db runner and exec runner.
For example, it can wait when updating values to the database asynchronously.
It is also possible to use it in the include runner to wait.

k2tzumi: Sure, there's asynchronous processing.

eval.go
```diff
@@ -26,3 +26,15 @@ func evalCond(cond string, store map[string]interface{}) (bool, error) {
...
+
+func evalCount(count string, store map[string]interface{}) (int, error) {
+	r, err := expr.Eval(count, store)
+	if err != nil {
+		return 0, err
+	}
+	c, ok := r.(int)
+	if !ok {
+		return 0, fmt.Errorf("invalid count: %v", count)
+	}
```
k2tzumi, line 38: [IMO] Since the result of `expr.Eval` is of multiple types, it would be better to convert it to int appropriately.

```suggestion
	var c int
	switch v := r.(type) {
	case string:
		c, _ = strconv.Atoi(v)
	case int64:
		c = int(v)
	case float64:
		c = int(v)
	case int:
		c = v
	default:
		return 0, fmt.Errorf("invalid count: evaluated %s, but got %T(%v)", count, r, r)
	}
```

loop.go
```diff
@@ -12,7 +13,9 @@ import (
 )
 
 const (
-	retrySectionKey = "retry"
+	loopSectionKey            = "loop"
+	deprecatedRetrySectionKey = "retry" // deprecated
+	loopCountVarKey           = "i"
```
k2tzumi, line 18: [nits]
How to name the key if we try to implement a counter for the index of steps[*] in a future feature addition? I was wondering.

Is there a way to concisely describe a case that references an index when testing a response in the same step?

```yml
steps:
  -
    req:
      /get?var=foo:
        get:
          body: null
    test: |
      steps[0].res.status == 200
```

I wanted to be concise, e.g., to be able to refer to the index of `steps[0]` by a variable.

eval.go
k1LoW: Thank you!!

loop.go
k1LoW: I have the same problem as you (To solve this problem, I created a step listing with a [map](https://github.com/k1LoW/runn#runbook--runn-scenario-file-).

> How to name the key if we try to implement a counter for the index of steps[*] in a future feature addition? I was wondering.

I'm certainly wondering what to do about it.

If we were to support both Array and Map types, should there be a way to specify the current step as follows?

```yaml
steps:
  -
    req:
      /get?var=foo:
        get:
          body: null
    test: |
      current.res.status == 200
```

```yaml
steps:
  reqget:
    req:
      /get?var=foo:
        get:
          body: null
    test: |
      current.res.status == 200
```

k1LoW: If you have any ideas for naming the step[*] counter and the loop counter well, please let me know!

k2tzumi: > If we were to support both Array and Map types, should there be a way to specify the current step as follows?

The `current` key looks very nice, since it can be used in either Map or Array.

I'd like to respond separately from this PR.

k1LoW: @k2tzumi Thank you very much!!

operator.go
```diff
@@ -811,32 +831,50 @@ func (o *operator) runInternal(ctx context.Context) error {
...
 					}
-					if tf {
-						success = true
-						break
+					if s.loop.Until != "" {
+						store := o.store.toMap()
+						t = buildTree(s.loop.Until, store)
+						o.Debugln("-----START RETRY CONDITION-----")
+						o.Debugf("%s", t)
+						o.Debugln("-----END RETRY CONDITION-----")
```
k2tzumi, line 861: [nits]
```suggestion
						o.Debugln("-----START LOOP CONDITION-----")
						o.Debugf("%s", t)
						o.Debugln("-----END LOOP CONDITION-----")
```

loop.go
k1LoW: @k2tzumi I invited you to become a runn collaborator!

operator.go
k1LoW: Thank you!!

k1LoW:
@k2tzumi If you can, I'd be happy to get your comments on whether or not it looks good for us.

testdata/book/httpbin.yml
```diff
@@ -33,28 +36,60 @@ steps:
...
+    req:
+      /status/{{ 204 - i }}:
+        get:
+          body: null
+    test: |
+      steps[5].res.status == (204 - i)
+  -
+    test: |
+      steps[5].res.status == 204
+    # FIXME: I'd like to see the last status be 200.
```
k2tzumi, line 87: @k1LoW 

What is the best rule for referencing the result of a loop process?

* Keep only the final result.
Currently, I think the first result is referenced.
* Allow reference to the result of each loop.
`steps[5].res[4].status`.

testdata/book/httpbin.yml
```diff
@@ -33,28 +36,60 @@ steps:
...
+  -
+    loop:
+      count: len(vars.loopVars)
+    include:
+      path: httpbin_include.yml
+      vars:
+        jsonRequestBody: 'json://{{ vars.loopVars[i] }}'
```
k2tzumi, line 94: I was expecting the json to be loaded. 😅 

I will try to respond as soon as possible.

testdata/book/httpbin_include.yml
```diff
@@ -0,0 +1,19 @@
...
+    endpoint: ${HTTPBIN_END_POINT:-https://httpbin.org/}
+vars:
+  jsonRequestBody:
+    foo: "test"
+    bar: 1
+  counter: 0
+steps:
+  -
+    req:
+      /post?count={{ vars.counter }}:
```
k2tzumi, line 13: [NOTE]
I thought I could reference the loop counter directly in the include destination, but I couldn't.
I don't think it's a problem because there are alternatives as per the test scenario.
```suggestion
      /post?count={{ i }}:
```

testdata/book/httpbin.yml
k1LoW: Maybe the until: condition is incorrect.

```diff
- until: 'steps[5].res.status != "200"'
+ until: 'steps[5].res.status == "200"'
```

testdata/book/httpbin_include.yml
k1LoW: > I thought I could reference the loop counter directly in the include destination, but I couldn't.

Yes. Because the location of the `i` variable is root, I thought it would conflict with the loop counter in the included.

I also thought `parent.i` would be uncool 😄

testdata/book/httpbin.yml
k1LoW: Great!!!

testdata/book/httpbin.yml
k1LoW: Fixed 👍 https://github.com/k1LoW/runn/pull/97/commits/c26f27c2243aafaaa48dea72387aacb643553615

k2tzumi (approved):
@k1LoW 
LGTM!

k1LoW:
Thank you!!!!

testdata/book/httpbin.yml
k2tzumi: PR sent.
https://github.com/k1LoW/runn/pull/101

PR #100, opened 2022-08-20
[TEST] loop: section

k2tzumi:
refs https://github.com/k1LoW/runn/pull/97

# Verification

- Ability to set loop counter from vars.
- Ability to use indexes in requests and tests in loop processing
- Ability to replace the retry process with a loop
- Ability to perform loop processing in combination with include

# What has not been verified

- To be retried when HTTP request times out
Error with `DeadlineExceeded`. May not have been handled originally
- The ability to refer to the results of an iterative process in a separate step.
- Ability to read external json using a loop counter.
#93 will be omitted from the response.

### DIFF
--- a/testdata/book/httpbin.yml
+++ b/testdata/book/httpbin.yml
@@ -17,13 +17,13 @@ vars:
       foo: "test2"
       bar: 2
   external: json://testdata/vars.json
+  loopCount: 3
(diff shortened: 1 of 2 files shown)


k1LoW:
Thank you !!!

PR #106, opened 2022-08-21
Append compare function

k2tzumi:
# Purpose

Exclude some Entries when validating the response body so that it can be compared with vars.

* Example
  ```yml
    -
      desc: All the same except Content-Length
      req:
        /response-headers?freeform={{ vars.wantBody.freeform }}:
          get:
            body: null
      test: |
        steps[8].res.status == 200
        && compare(steps[8].res.body, vars.wantBody, "Content-Length")
  ```

# Improvement details

* Define `go-cmp/cmp` compare functions as built-in functions
Is the selection of the library to be used reasonable?

# Issue

* debug output is not optimized for syntax with functions
  ```
  -----START TEST CONDITION-----
  steps[8].res.status == 200
  && compare(steps[8].res.body, vars.wantBody, "Content-Length")
  
  
  ├── steps[8].res.status => 200
  ├── 200 => 200
  ├── compare(steps[8].res.body => ?
  ├── vars.wantBody => {"Content-Type":"application/json","freeform":"foo"}
  ├── "Content-Length" => "Content-Length"
  └── ) => ?
  -----END TEST CONDITION-----
  ```
* Diff contents are not output to debug
It might be better to use a built-in function to determine debug flags and output them directly
* Specifying which Entries to exclude is only supported for the first level.
* ~~Not a concise evaluation formula.~~

### DIFF
--- a/option.go
+++ b/option.go
@@ -357,16 +357,20 @@ func setupBuiltinFunctions(opts ...Option) []Option {
 		Func("string", func(v interface{}) string { return cast.ToString(v) }),
 		Func("int", func(v interface{}) int { return cast.ToInt(v) }),
 		Func("bool", func(v interface{}) bool { return cast.ToBool(v) }),
-		Func("diff", cmp.Diff),
-		Func("ignoreMapEntries", func(ignoreKeys ...string) cmp.Option {
-			return cmpopts.IgnoreMapEntries(func(key string, val interface{}) bool {
+		Func("compare", func(x, y interface{}, ignoreKeys ...string) bool {
+			diff := cmp.Diff(x, y, cmpopts.IgnoreMapEntries(func(key string, val interface{}) bool {
 				for _, ignore := range ignoreKeys {
 					if key == ignore {
 						return true
 					}
 				}
 				return false
-			})
+			}))
+
+			// FIXME: Debug output of diffs
+
+			return diff == ""
+
 		}),
 	},
 		opts...,
--- a/option.go
+++ b/option.go
@@ -14,6 +14,8 @@ import (
 	"time"
 
 	"github.com/bmatcuk/doublestar/v4"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/jhump/protoreflect/desc"
 	"github.com/spf13/cast"
 	"google.golang.org/grpc"
@@ -355,6 +357,17 @@ func setupBuiltinFunctions(opts ...Option) []Option {
 		Func("string", func(v interface{}) string { return cast.ToString(v) }),
 		Func("int", func(v interface{}) int { return cast.ToInt(v) }),
 		Func("bool", func(v interface{}) bool { return cast.ToBool(v) }),
+		Func("diff", cmp.Diff),
+		Func("ignoreMapEntries", func(ignoreKeys ...string) cmp.Option {
+			return cmpopts.IgnoreMapEntries(func(key string, val interface{}) bool {
+				for _, ignore := range ignoreKeys {
+					if key == ignore {
+						return true
+					}
+				}
+				return false
+			})
+		}),
 	},
 		opts...,
 	)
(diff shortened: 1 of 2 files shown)


k1LoW:
I think it's a very interesting idea!

What we would like to provide is a means to simplify the comparison of complex data such as response bodies.

>it can be **compared with** vars

> It may be better to wrap the original diff function and implement the **assertEquals** function.

Considering the use case, I too think it might be better to provide it as `compare(a, b interface{}) bool` or `compare(a, b interface{}) bool`.

> Exclude some Entries

I do not want the exclusion function to be a cause of increased complexity.

Many comparison functions for testing do not have an exclusion function.

However, I want the exclusion feature too 🌗

k2tzumi:
Thank you for the ideas!

> Considering the use case, I too think it might be better to provide it as compare(a, b interface{}) bool or compare(a, b interface{}) bool.

I still prefer bool for return values.

> I do not want the exclusion function to be a cause of increased complexity.

Maybe allow exclusivity to be specified as an optional argument?

I noticed the following notation in the test syntax

https://github.com/k1LoW/runn/blob/main/test_test.go#L38-L40

`A in B`
`A contains B`

How is this feature implemented?

Would it be easier to evaluate readability and expressions if the notation were provided as a DSL?

For example
`match A and B exclude "foo", "bar"`

Tools that allow you to write scenarios in DSL include

https://github.com/karatelabs/karate

Gherkin is the most famous notation.

https://cucumber.io/docs/gherkin/

k1LoW:
> How is this feature implemented?

It is the syntax of the built-in evaluation engine.

https://github.com/k1LoW/runn#expression-evaluation-engine

I am currently reluctant to have our own DSL.

k1LoW:
> Maybe allow exclusivity to be specified as an optional argument?

Seems like a good option to me.

option.go
```diff
@@ -355,6 +357,21 @@ func setupBuiltinFunctions(opts ...Option) []Option {
...
+			diff := cmp.Diff(x, y, cmpopts.IgnoreMapEntries(func(key string, val interface{}) bool {
+				for _, ignore := range ignoreKeys {
+					if key == ignore {
+						return true
+					}
+				}
+				return false
+			}))
+
+			// FIXME: Debug output of diffs
```
k2tzumi, line 370: At this point, it was possible to refer to the debug of the book object, but because the operator did not exist, debug output could not be generated.

option.go
```diff
@@ -355,6 +357,21 @@ func setupBuiltinFunctions(opts ...Option) []Option {
 		Func("string", func(v interface{}) string { return cast.ToString(v) }),
 		Func("int", func(v interface{}) int { return cast.ToInt(v) }),
 		Func("bool", func(v interface{}) bool { return cast.ToBool(v) }),
+		Func("compare", func(x, y interface{}, ignoreKeys ...string) bool {
```
k2tzumi, line 360: Is this the correct interface for the compare function?

k2tzumi:
> debug output is not optimized for syntax with functions

Fixing debugging output would require revisiting lexical analysis, which would be more difficult

option.go
k1LoW: Looks good to me!

k1LoW:
> Fixing debugging output would require revisiting lexical analysis, which would be more difficult

I think we can consider this as a known issue at another time.

option.go
k1LoW: Just a thought, how about adding `diff()` too?

I could use `diff()` with Dump runner, maybe for debugging `compare()`.

k2tzumi: https://github.com/k1LoW/runn/issues/112

It seems like a good idea to have a diff function once the above support is complete.

```
-----START TEST CONDITION-----
diff(steps[8].res.body, vars.wantBody, "Content-Length") != ""


├── diff(steps[8].res.body, vars.wantBody, "Content-Length")
｜         ├── steps[8].res.body => {"Content-Type":"application/json","freeform":"foo", "Content-Length": 81}
｜         ├── vars.wantBody => {"Content-Type":"application/json","freeform":"foo"}
｜         └── "Content-Length" => "Content-Length"
└──  (diff contents) => ""
-----END TEST CONDITION-----
```

k1LoW: GREAT!!!!

k1LoW (approved):
LGTM!

Issue #113, opened 2022-08-25
Ensure a means of accessing the current step variable


k2tzumi:
refs https://github.com/k1LoW/runn/pull/97#discussion_r950659660

# Proposal

If the request and test are executed in the same step, we want easy access to the response results.

Define the `current` variable so that it can be accessed by the same variable name in the named and unnamed steps.

```yaml
steps:
  -
    req:
      /get?var=foo:
        get:
          body: null
    test: |
      current.res.status == 200
```

```yaml
steps:
  reqget:
    req:
      /get?var=foo:
        get:
          body: null
    test: |
      current.res.status == 200
```

k1LoW:
Looks good to me!!

PR #119, opened 2022-08-27
Support `current` variable

k1LoW:
Fix https://github.com/k1LoW/runn/issues/113

### DIFF
--- a/test.go
+++ b/test.go
@@ -24,8 +24,11 @@ func newTestRunner(o *operator) (*testRunner, error) {
 	}, nil
 }
 
-func (rnr *testRunner) Run(ctx context.Context, cond string) error {
+func (rnr *testRunner) Run(ctx context.Context, cond string, runned bool) error {
 	store := rnr.operator.store.toMap()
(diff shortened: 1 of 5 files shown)


test.go
```diff
@@ -24,8 +24,11 @@ func newTestRunner(o *operator) (*testRunner, error) {
 	}, nil
 }
 
-func (rnr *testRunner) Run(ctx context.Context, cond string) error {
+func (rnr *testRunner) Run(ctx context.Context, cond string, runned bool) error {
 	store := rnr.operator.store.toMap()
+	if runned {
+		store[storeCurrentKey] = rnr.operator.store.latest()
+	}
```
k2tzumi, line 31: I see, indeed, that’s right

So you are enabling the current variable only after execution?

k2tzumi (approved):
Thanks a lot!
LGTM!

test.go
k1LoW: 👍 I think this was the only point in time to use a current variable.

Issue #112, opened 2022-08-25
Optimize debugging output of function evaluation expressions


k2tzumi:
# Proposal

Debug output in the test syntax is the following output when using functions

* As-Is
  ```
  -----START TEST CONDITION-----
  compare(steps[8].res.body, vars.wantBody, "Content-Length")
  
  
  ├── compare(steps[8].res.body => ?
  ├── vars.wantBody => {"Content-Type":"application/json","freeform":"foo"}
  ├── "Content-Length" => "Content-Length"
  └── ) => ?
  -----END TEST CONDITION-----
  ```

I want to optimize the output as follows

* To-be
  ```
  -----START TEST CONDITION-----
  compare(steps[8].res.body, vars.wantBody, "Content-Length")
  
  
  └── compare(steps[8].res.body, vars.wantBody, "Content-Length") => true
              ├── steps[8].res.body => {"Content-Type":"application/json","freeform":"foo", "Content-Length": 81}
              ├── vars.wantBody => {"Content-Type":"application/json","freeform":"foo"}
              └── "Content-Length" => "Content-Length"
  -----END TEST CONDITION-----
  ```

k1LoW:
GREAT!!!

While your suggestion is better I think, I believe the following indentation would be useful enough.

  ```
  -----START TEST CONDITION-----
  compare(steps[8].res.body, vars.wantBody, "Content-Length")
  

  ├── compare(steps[8].res.body, vars.wantBody, "Content-Length") => true
  ├── steps[8].res.body => {"Content-Type":"application/json","freeform":"foo", "Content-Length": 81}
  ├── vars.wantBody => {"Content-Type":"application/json","freeform":"foo"}
 └── "Content-Length" => "Content-Length"
  -----END TEST CONDITION-----
  ```

k1LoW:
I believe the format of https://github.com/k1LoW/runn/issues/112#issuecomment-1227271392 could be done with #120.

k2tzumi:
Thanks for the great PR!

PR #120, opened 2022-08-28
Use parser.Parse() instead of own parser

k1LoW:
ref: #112


k1LoW:
I found that some cases were not considered, so I am committing them additionally

k1LoW:
Fix cases of https://github.com/antonmedv/expr/blob/master/docs/Language-Definition.md#supported-literals

k2tzumi (approved):
Looks like additional testing is needed, but it looked good!

The extended syntax here can't be parsed by AST, can it?
https://github.com/antonmedv/expr/blob/master/docs/Language-Definition.md

k1LoW:
> The extended syntax here can't be parsed by AST, can it?

It probably cannot be parsed. The current runn built-in function can be parsed as a function.

k2tzumi:
> It probably cannot be parsed. The current runn built-in function can be parsed as a function.

As I thought.
Most of the syntax is supported, so it should not be a problem 👍

PR #127, opened 2022-08-30
Support uint64 expand

k2tzumi:
# Problem

Errors were occurring when trying to process uint64 during variable expansion.

```
invalid format: evaluated vars.uint64 , but got uint64(4600)
```

Let uint64 support.

I was able to verify this on a unit test level, but could not reproduce the error when run through RunBook.
I reproduced this when using the brew version, but I wonder if it has anything to do with it?

### DIFF
--- a/operator.go
+++ b/operator.go
@@ -917,6 +917,8 @@ func (o *operator) expand(in interface{}) (interface{}, error) {
 				}
 			case int64:
 				s = strconv.Itoa(int(v))
+			case uint64:
+				s = strconv.Itoa(int(v))
 			case float64:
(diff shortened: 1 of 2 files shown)


k2tzumi:
I'd like to think there are no more leaks elsewhere 😅

k1LoW (approved):
> I reproduced this when using the brew version, but I wonder if it has anything to do with it?

Now I don't know why either.

> I was able to verify this on a unit test level, but could not reproduce the error when run through RunBook.

Reproduction with only test cases will be enough!

Issue #124, opened 2022-08-30
Error when selecting date column in dBRunner


k2tzumi:
# Problem

The following error occurred.

```
strconv.Atoi: parsing “2022-08-30 15:07:26”: invalid syntax
```

Verified DB was MySQL

k2tzumi:
refs https://github.com/k1LoW/runn/pull/128#issuecomment-1232302966

PR #129, opened 2022-08-31
Support for DATETIME columns in DBRunner

k2tzumi:
fix https://github.com/k1LoW/runn/issues/124

# Reason for concern

* Isn't the DATETIME format different for different types of databases?
Can you identify the format from the DB connection driver?
No generic DateTime formatting library exists?
* What time zone should the date be in?
* How to implement the verification code?
* Support other formats besides DATETIME?
The following types exist, but they are all treated as numbers
DECIMAL, BOOL, INT, BIGINT

### DIFF
--- a/db.go
+++ b/db.go
@@ -92,7 +92,7 @@ func (rnr *dbRunner) Run(ctx context.Context, q *dbQuery) error {
 						s := string(v)
 						t := strings.ToUpper(types[i].DatabaseTypeName())
 						switch t {
-						case "TEXT", "VARCHAR", "NVARCHAR":
+						case "TEXT", "CHAR", "VARCHAR", "NVARCHAR":
 							row[c] = s
 						case "DATETIME":
 							dt, err := time.Parse("2006-01-02 15:04:05", s)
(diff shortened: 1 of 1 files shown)


k1LoW (approved):
LGTM! Thank you!

I understand your concern.
The handling of return values will need to be improved in the future.

k2tzumi:
I tried running it locally and got an error.
It seems that the layout is not well specified.

```
invalid datetime column: evaluated created_at, but got DATETIME(2022-09-01 19:07:30): parsing time "2016-01-02 15:04:05" as "2022-09-01 19:07:30": cannot parse "-01-02 15:04:05" as "2"
```

db.go
```diff
@@ -95,7 +95,7 @@ func (rnr *dbRunner) Run(ctx context.Context, q *dbQuery) error {
 						case "TEXT", "VARCHAR", "NVARCHAR":
 							row[c] = s
 						case "DATETIME":
-							dt, err := time.Parse(s, "2006-01-02 15:04:05")
+							dt, err := time.Parse("2006-01-02 15:04:05", s)
```
k2tzumi, line 98: The specified argument was incorrect. 🤮

Issue #117, opened 2022-08-27
Support embedding variables in json files


k2tzumi:
# Proposal

To allow variables to be embedded in json files for more flexibility as test data.

```json
{
    "code": "{{ vars.code }}",
    "data": {
       "foo": "test",
      "bar": 1
   }
}
```

Use vars syntax to expand variables when reading json.

k1LoW:
Very interesting 👍 

I think the following implementation might be difficult
- Expansion of Integer and Boolean types

k2tzumi:
> Expansion of Integer and Boolean types

I think I was limited to only within a string to expand a variable, so this problem is difficult.

Other tools seemed to have the same problem. 😓 
https://community.postman.com/t/the-path-variable-var-needs-to-be-of-type-integer-but-we-found-var/12714

PR #132, opened 2022-09-03
Build vars parameters with templates

k2tzumi:
close https://github.com/k1LoW/runn/issues/117

# What we did

When a json file is read, if the file name is t `.json.template`, the template engine is applied and the parameters are bound to vars.
This modification only supports vars override for include syntax.


testdata/template.json
```diff
@@ -0,0 +1,4 @@
+{
+    "foo": "{{.vars.foo -}}",
+    "bar": {{.vars.bar }}
```
k2tzumi, line 3: Since it is a template file, I think I can ignore json syntax errors.

testdata/template.json
```diff
@@ -0,0 +1,4 @@
+{
+    "foo": "{{.vars.foo -}}",
```
k1LoW, line 2: I think it is good that JSON is first determined as a string and then variables are evaluated.

However, I am concerned about the new Go template syntax that has been implemented.

I' d like to have one syntax if possible.

k1LoW: JSON and runn syntax are not a good match...

k2tzumi: > However, I am concerned about the new Go template syntax that has been implemented.
> 
> I' d like to have one syntax if possible

* `antonmedv/expr`
https://github.com/antonmedv/expr/blob/master/docs/Language-Definition.md
* `text/template`
https://pkg.go.dev/text/template#hdr-Text_and_spaces

Would `antonmedv/expr` be better?
The syntax looks almost the same, but `antonmedv/expr` has more user-friendly error output and object-based I/F, so is it easier to handle?

k1LoW: As a user of runn, I do not want to simplify its user interface.

I believe that one can choose not to implement a feature if it is to avoid complexity.

> Would antonmedv/expr be better?

I don't know if it is better, but at least runn has already introduced this I/F.

k2tzumi: I agree with you about keeping it simple with one syntax.

> I don't know if it is better, but at least runn has already introduced this I/F.

There is no positive reason to review the syntax we have adopted for this feature addition, so we will go with `antonmedv/expr`.

k1LoW: Thank you...!!

k2tzumi: https://pkg.go.dev/text/template#hdr-Actions

I've noticed that I can use the range syntax.
It's gotten greedy in my mind.

Here is a summary of the interests that each library and what we hope to achieve as a runn.

  | antonmedv/expr | text/template
-- | -- | --
Domain interest to be handled | Evaluate the expression | Provide template functions
Scope of Application | runbook(yaml) | Request and expected data(json)
Timing of Usage | When evaluating expressions in the syntax | When reading json
Restrictions on Use | Used only for values on yaml.Do not let runn syntax be dynamically generated | No restrictions.
Required Flexibility | Not so important. | Considered to be important

I felt that the use cases were different when organized.
I realized that I wanted a template engine to handle requests and expectations.
Conversely, we felt that it was undesirable to unify the syntax of evaluation expressions, because too much freedom would break the syntax of the runbook!

I think the concern about high cognitive load in introducing a new syntax is somewhat acceptable by using Go's standard template engine, text/template.

We envisioned the following specific use cases where you would want to use the template engine.

* displaying a list screen and performing batch processing.

When creating a scenario with the above use case, you will need to add a repeating element to the request data based on the results of the list fetch API.
If we try to achieve this using evaluation expressions, we must make good use of the builtin functions
Regardless of the above use cases, there are many cases where request data is processed in a complex manner on the front end in today's SPA.
We believe that request data is easier to manage if it is created from scratch using a template engine.

At first glance, the new syntax may seem to add complexity, but we thought it was an important element to simplify the scenario.

We believe that this can be accomplished to some extent by combining evaluation formulas and built-in functions, but we are concerned that many modifications will be necessary to accommodate increasingly complex data.

I know this is a long post, but I would appreciate your input!

k1LoW: Honestly, I can't imagine a use case for the functionality that this Pull Request would provide.
So maybe I don't want to complicate it.

I am also concerned about the similar but different syntax of text/template and evaluation expressions.
Because we believe that runn should be a tool that can be used for more than just Gopher (certainly Gopher would not be confused.).
I think the similar but different syntax is confusing.

On the other hand, I agree with what you are saying.

In other words, we may not know until we use it.

This is just a suggestion, how about introducing it on a trial implementation basis?

---

> [Scope of Application] Request and expected data(json)

I believe it is possible that I am not limited to this.

> use the template engine

If there were a mechanism to receive the results of expansion by an external command, it would be much more flexible and runn would be less complicated.

k1LoW: It might be nice to have a mechanism where text/template works depending on the file name.

For example 

- `path/to/somerequest.json` ... normal json
- `path/to/somerequest.json.template` ... text/template works

k2tzumi: > If there were a mechanism to receive the results of expansion by an external command, it would be much more flexible and runn would be less complicated.

I had not thought of that idea.
If we can separate the scenario and test data in a nice way, and if we can smoothly link data with external tools, we can keep the run implementation and scenario simple.

> It might be nice to have a mechanism where text/template works depending on the file name.

I like this idea too. Explicitly shows the use of the template.

> In other words, we may not know until we use it.
>
> This is just a suggestion, how about introducing it on a trial implementation basis?

Thank you.
We'd like to dogfood and explore better ideas.

k2tzumi: > It might be nice to have a mechanism where text/template works depending on the file name.

Added condition to apply template by file name
https://github.com/k1LoW/runn/pull/132/commits/a6ffec8d7214235020147dac569ef931343ab794

vars.go
```diff
@@ -1,23 +1,36 @@
...
 	case string:
 		if strings.HasPrefix(v, varsSupportScheme) {
-			byteArray, err := os.ReadFile(v[len(varsSupportScheme):])
+			fn := v[len(varsSupportScheme):]
+			byteArray, err := os.ReadFile(fn)
 			if err != nil {
 				return value, fmt.Errorf("read external files error: %w", err)
-
+			}
+			if store != nil && strings.HasSuffix(fn, ".json.template") {
```
k1LoW, line 24: I think `".json.template"` should be const.

k1LoW (approved):
NICE!! Thank you!!

PR #134, opened 2022-09-05
Improve column type determination using MySQL integration testing environment

k1LoW:
Improve column type determination of DB Runner using MySQL integration testing environment.

Also, in the built-in function, I added `time()` which casts the string to time.Time{}.


db.go
```diff
@@ -91,24 +91,30 @@ func (rnr *dbRunner) Run(ctx context.Context, q *dbQuery) error {
...
+						case strings.Contains(t, "TEXT") || strings.Contains(t, "CHAR") || t == "TIME": // MySQL8: ENUM = CHAR
 							row[c] = s
-						case "DATETIME":
+						case t == "DECIMAL" || t == "FLOAT" || t == "DOUBLE": // MySQL: NUMERIC = DECIMAL
+							num, err := strconv.ParseFloat(s, 64)
+							if err != nil {
+								return fmt.Errorf("invalid column: evaluated %s, but got %s(%v): %w", c, t, s, err)
+							}
+							row[c] = num
+						case t == "DATETIME": // TODO: Need test case
```
k1LoW, line 103: @k2tzumi If possible, can you tell me the database where the following URL error occurred?

https://github.com/k1LoW/runn/pull/129#issuecomment-1234057949

k2tzumi: The following versions and images were verified
`public.ecr.aws/docker/library/mysql:8.0.29`

k1LoW: Thank you!

option.go
```diff
@@ -13,6 +13,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/araddon/dateparse"
```
k2tzumi, line 16: I didn't know there was such a useful library. It's helpful!

db.go
```diff
@@ -91,24 +91,30 @@ func (rnr *dbRunner) Run(ctx context.Context, q *dbQuery) error {
 					case []byte:
 						s := string(v)
 						t := strings.ToUpper(types[i].DatabaseTypeName())
-						switch t {
-						case "TEXT", "CHAR", "VARCHAR", "NVARCHAR":
+						switch {
+						case strings.Contains(t, "TEXT") || strings.Contains(t, "CHAR") || t == "TIME": // MySQL8: ENUM = CHAR
```
k2tzumi, line 95: Thank you for getting this straightened out!

db.go
k1LoW: @k2tzumi  Maybe this case statement is not working in the code.

https://github.com/k1LoW/runn/pull/135#issuecomment-1236901855

Can I try to turn it off in this Pull Request?

k2tzumi: https://github.com/k1LoW/runn/releases/tag/v0.32.2

I verified the above tag with the following error

```
invalid column: evaluated created_at, but got DATETIME(2022-09-05 22:46:41): strconv.Atoi: parsing "2022-09-05 22:46:41": invalid syntax
```

It worked on the local environment image `mysql:8.0.29`
Maybe there is a subtle difference from the image of AWS..

> Can I try to turn it off in this Pull Request?

We have not yet incorporated this into the actual scenario, so you can turn it off.

k2tzumi: Oops? I tried to match the local image, but it worked on this branch too!

```
-----START ROWS-----
+-----------+-------------------------------+-------------------------------+
| is_locked |          created_at           |          updated_at           |
+-----------+-------------------------------+-------------------------------+
|         1 | 2022-09-05 23:01:16 +0000 UTC | 2022-09-05 23:01:16 +0000 UTC |
+-----------+-------------------------------+-------------------------------+
(1 row)
-----END ROWS-----
```

On CI, there may be a problem with locale combinations.

k1LoW: > We have not yet incorporated this into the actual scenario, so you can turn it off.

Thank you.
I will delete the CASE statement once and wait for the problem you encountered to occur.

PR #140, opened 2022-09-11
Expand vars of Include while preserving type information

### LINKED ISSUE #139: Improved vars takeover when using multi-level includes
# Problem

Inability to successfully take over vars when included at multiple levels

* a.yml
```yaml
vars:
   foo: 123
steps:
  -
   include: b.yml
   vars:
     foo: '{{ vars.foo }}'

.. snip ..
```
* b.yml
```yaml
vars:
   foo: 123
steps:
  -
   include: c.yml
   vars:
     foo: '{{ vars.foo }}'

.. snip ..
```
* c.yml
```yaml
vars:
   foo: 123
steps:
  -
    test: vars.foo == 123
```

Error when executing a.yml
The value of foo is the string `'123'`.

# Question

If I want to specify overwriting the vars in `c.yml` from `a.yml` in the above example, is the above description the only way?
Is there any problem to make the specification that vars defined in the parent yaml are inherited?
When specifying vars at include time, I feel that numerical values are inevitably evaluated as strings.

k2tzumi:
fix #139

### DIFF
--- a/testdata/book/multiple_include_a.yml
+++ b/testdata/book/multiple_include_a.yml
@@ -7,4 +7,4 @@ steps:
       path: multiple_include_b.yml
       vars:
         foo: '{{ vars.foo }}'
-    test: vars.foo == '123'
+    test: vars.foo == 123
--- a/testdata/book/multiple_include_a.yml
+++ b/testdata/book/multiple_include_a.yml
@@ -4,6 +4,7 @@ vars:
   bar: "123-abc"
   baz: "-23"
   qux: 4
+  quux: "2"
 steps:
   b:
     include:
@@ -13,4 +14,5 @@ steps:
         bar: '{{ vars.foo }}-{{ vars.bar }}'
         baz: '-{{ vars.foo - len(vars.bar) }}'
         qux: '{{ len(vars.bar) - vars.qux }}'
-    test: vars.foo == 123 && vars.bar == "123-abc" && vars.baz == "-23" && vars.qux == 4
+        quux: '{{ string(int(vars.quux) + vars.qux) }}'
+    test: vars.foo == 123 && vars.bar == "123-abc" && vars.baz == "-23" && vars.qux == 4 && vars.quux == "2"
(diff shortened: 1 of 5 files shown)


k2tzumi:
@k1LoW 

Can you give me your opinion on the implementation image?
May need refactoring, including the [expand](https://vscode.dev/github/k1LoW/runn/blob/f343c14dbe039775078afd46f85ddfd0b7ce38b0/operator.go#L899) function

testdata/book/multiple_include_a.yml
```diff
@@ -0,0 +1,18 @@
+desc: For multiple include test
+vars:
+  foo: 123
+  bar: "123-abc"
+  baz: "-23"
+  qux: 4
+  quux: "2"
```
k1LoW, line 7: How about adding `if: included` to multiple_include_a.yml and multiple_include_b.yml so we can better understand the behavior of override vars?

testdata/book/multiple_include_a.yml
```diff
@@ -0,0 +1,18 @@
...
+  b:
+    include:
+      path: multiple_include_b.yml
+      vars:
+        foo: '{{ vars.foo }}'
+        bar: '{{ vars.foo }}-{{ vars.bar }}'
+        baz: '-{{ vars.foo - len(vars.bar) }}'
+        qux: '{{ len(vars.bar) - vars.qux }}'
+        quux: '{{ string(int(vars.quux) + vars.qux) }}'
+    test: vars.foo == 123 && vars.bar == "123-abc" && vars.baz == "-23" && vars.qux == 4 && vars.quux == "2"
```
k1LoW, line 18: Ideally, we would like `vars.quux == 2` instead of `vars.quux == "2"` ( I understand how difficult it is ).

k1LoW:
GREAT WORK!!! I think it is clear that it is at least better than what we have now.

k1LoW (approved):
LGTM!

k2tzumi:
I tested it by integrating it into the actual scenario and found another error.
Please wait a moment while we investigate.

k2tzumi:
Maybe this will work.
https://github.com/k1LoW/runn/pull/140/commits/e8cf3972399df62d8dadb12acf211290c7e27973

k1LoW (approved):
Thank you!!

k1LoW:
I will release a new version at least tomorrow!

Issue #103, opened 2022-08-21
TestShard is flaky


k1LoW:
TestShard is flaky

ref: https://github.com/k1LoW/runn/pull/95#issuecomment-1214261446

k2tzumi:
refs https://github.com/k1LoW/runn/pull/128#issuecomment-1232290831

k1LoW:
refs: https://github.com/goccy/go-yaml/issues/316

k1LoW:
refs: https://github.com/k1LoW/runn/actions/runs/3068252022/jobs/4955490730#step:6:77

😿

k1LoW:
I think we''ve solved this problem, so Close.

PR #147, opened 2022-09-16
[BREAKING] Use gopkg.in/yaml.v2 instead

k1LoW:
ref: https://github.com/k1LoW/runn/issues/103

goccy/go-yaml keeps the code simpler but currently has unstable issues.

So we use gopkg.in/yaml.v2 instead.


k1LoW:
We believe this will be a temporary fix.

yaml.go
```diff
@@ -0,0 +1,58 @@
...
+
+// Mainly respond to issue where the deep nested map becomes map[interface{}]interface{}
+// and issue where yaml.MapSlice is defined as nested.
+
+func yamlMarshal(in interface{}) ([]byte, error) {
+	return yaml.Marshal(in)
+}
+
+func yamlUnmarshal(in []byte, out interface{}) error {
+	return yaml.Unmarshal(in, out)
```
k2tzumi, line 17: @k1LoW 
Is there any reason why you didn't have it normalized within the `yamlUnmarshal` function?

k1LoW: For the following reasons

- I didn't want to change the function signatures because I plan to revert to goccy/go-yaml
- On the other hand, the type of `out interface{}` was undefined, so it was difficult to implement without changing the signature

If the above can be resolved, it would be better to normalize within yamlUnmarshal as you say.

By the way, I created yamlUnmarshal to collect concrete implementations of YAML operations in yaml.go.
As a result, yamlUnmarshal and yamlUnmarshal are the same.

k1LoW: And, the reason I created the `normalize()` is to bridge as much as possible the difference between goccy/go-yaml and go-yaml/yaml's Unmarshal function.

k2tzumi: > I didn't want to change the function signatures because I plan to revert to goccy/go-yaml

Although the direct cause has not been identified, some vars are not being deployed properly from the version that merged this PR.

Some parts of the process may not be normalized.

If this PR is going to revert with a trial content, should I just sit tight for once?

k1LoW: > Although the direct cause has not been identified, some vars are not being deployed properly from the version that merged this PR.

That's not good.

Initially, this PR was created to resolve #103.

However, if it is not working well, we might revert this PR.

k1LoW: Or give up the revert to goccy/go-yaml...

k2tzumi: Or we can wait for the following Issue to be resolved
https://github.com/goccy/go-yaml/issues/316

k1LoW: @k2tzumi I have decided to revert this pr once.
#152

PR #150, opened 2022-09-18
Relative paths to json reads

### LINKED ISSUE #146: vars syntax to read json with relative paths
Currently it is loaded relative to the runn execution path, but I would like to load it relative to the scenario file.
We believe that the above specification is preferable when trying to manage multiple scenario files distributed in multiple directories.

k2tzumi:
close #146

### DIFF
--- a/operator.go
+++ b/operator.go
@@ -21,7 +21,6 @@ import (
 	"github.com/goccy/go-json"
 	"github.com/k1LoW/expand"
 	"github.com/k1LoW/stopw"
-	"github.com/rs/xid"
 	"go.uber.org/multierr"
 )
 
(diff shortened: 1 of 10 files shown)


operator.go
```diff
@@ -187,17 +186,12 @@ func New(opts ...Option) (*operator, error) {
...
-		o.id = bk.path
-		o.root = filepath.Dir(bk.path)
-	} else {
-		o.id = xid.New().String()
-		wd, err := os.Getwd()
-		if err != nil {
-			return nil, err
-		}
-		o.root = wd
+	o.id = bk.generateOperatorId()
```
k1LoW, line 189: Thank you!!

k1LoW (approved):
LGTM!!

PR #156, opened 2022-09-22
Move the parsing operation of runbook(YAML) from operator{} to book{} to clarify responsibilities

k1LoW:
Currently, the initial parsing (yaml.Unmarshal) of the runbook is performed by operator{} and book{}, so we will centralize it in book{}.

### DIFF
--- a/book.go
+++ b/book.go
@@ -21,21 +21,21 @@ import (
 const noDesc = "[No Description]"
 
 type book struct {
-	Desc          string                   `yaml:"desc,omitempty"`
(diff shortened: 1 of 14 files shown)


k2tzumi:
What would be the scope of function and structure exposure if runn were to act as a helper for go test?

k1LoW:
> What would be the scope of function and structure exposure if runn were to act as a helper for go test?

In this pull request, some fields that were unnecessarily `exported` are now `unexported`.

This does not affect the usage as a test helper.

PR #160, opened 2022-09-25
Change loop index variable name

k1LoW:


k2tzumi (approved):
The nest is deep, so I'd like to refactor it somewhere.

PR #163, opened 2022-09-26
Decimal point was not considered in the parsing of duration.

k1LoW:
Decimal point was not considered in the parsing of duration.

### DIFF
--- a/loop.go
+++ b/loop.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/goccy/go-yaml"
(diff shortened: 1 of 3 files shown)


loop.go
```diff
@@ -32,70 +33,91 @@ type Loop struct {
...
 	}
-	if r.Jitter == nil {
-		r.Jitter = &defaultJitter
+	if l.Jitter == nil {
+		l.Jitter = &defaultJitter
 	}
-	return r, nil
+
+	if l.Interval != "" {
+		i, err := parseDuration(l.Interval)
```
k1LoW, line 81: Add parsing error handling

Issue #145, opened 2022-09-14
Common settings for runners


k2tzumi:
# Proposal

Duplicate runner settings in each Runbook when creating multiple scenarios.
It would be good to define a runn configuration file and summarize in it the options that can be specified for runners and CLI execution.
Enables scenarios to be written in a reduced amount of detail and allows for separation of settings for each environment.

k1LoW:
I did not notice this because I use runn in my Go code.

In such cases, I prefer to create a `--overlay` ( `--underlay` ) option like [customize](customize).

- https://github.com/pepabo/control-controls#overlay
- https://github.com/k1LoW/tbls-build#architecture

The target to be overwritten, or the default value, does not have to be limited to `runners:`

PR #168, opened 2022-09-27
Improve post-processing of YAML parsed results

k1LoW:


k1LoW:
It's an improvement that will link to https://github.com/k1LoW/runn/issues/145#issuecomment-1246536240

PR #180, opened 2022-09-30
Show number of loop times when an error occurs during loop.

k1LoW:
Fix #178

PR #182, opened 2022-09-30
Fix buildTree()

k1LoW:
ref: https://github.com/k1LoW/runn/pull/180#issuecomment-1263707873

Fix the following incorrect displays

```
vars.tests[i] == vars.wants[i]
├──  => ?
└──  => ?
```


k2tzumi:
Thank you for fix the area I was concerned about!

k2tzumi (approved):

PR #183, opened 2022-10-01
Fixed compare function always returning true

k2tzumi:
I really wanted to test the built-in functions by themselves, but it was difficult, so I committed only the intermediate tests.

Can I define the built-in functions by themselves?
I want to check the behavior of the compare function.


k1LoW:
@k2tzumi Thank You!!!!

PR #185, opened 2022-10-01
migrate builtin functions to package

k2tzumi:
T/O

### DIFF
--- a/option.go
+++ b/option.go
@@ -2,7 +2,6 @@ package runn
 
 import (
 	"database/sql"
-	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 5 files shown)


option.go
```diff
@@ -467,46 +464,8 @@ func setupBuiltinFunctions(opts ...Option) []Option {
...
-				return false
-			}))
-
-			// FIXME: Debug output of diffs
-
-			return diff == ""
-
-		}),
+		Func("time", builtin.Time),
+		Func("compare", builtin.Compare),
```
k2tzumi, line 468: I want to automatically expand the public functions of the buildin package.

PR #342, opened 2022-12-02
Removed reset process for request Body for temporary support

k2tzumi:
# Triggers for Change

https://github.com/k1LoW/runn/releases/tag/v0.51.0

The previously addressed issue has reappeared in the above version.

https://github.com/k1LoW/runn/pull/91

The following library versions have changed due to the pkgs update

github.com/getkin/kin-openapi@v0.106.0 -> github.com/getkin/kin-openapi@v0.109.0

The contents of the release are as follows
https://github.com/getkin/kin-openapi/releases/tag/v0.109.0

# What we verified

1. If `skipValidateRequest: false`, no error occurred
2. Downgraded to `github.com/getkin/kin-openapi@v0.108.0` version and no more errors

# Cause Considerations

We believe the error recurred due to the following PR
https://github.com/getkin/kin-openapi/pull/672

The request body reset process in #91 is no longer needed due to modifications to the library itself.
On the contrary, the inclusion of provisional support seems to have caused the content length to be caught in the content length check.

# Changes

Deleted the process supported by #91 
My project is working well with this response.

# Worries

There may be an impact on the functionality of Multipart Form Data

### DIFF
--- a/http.go
+++ b/http.go
@@ -257,16 +257,13 @@ func (rnr *httpRunner) Run(ctx context.Context, r *httpRequest) error {
 		if err != nil {
 			return err
 		}
-		rc, ok := reqBody.(io.ReadCloser)
-		if !ok && reqBody != nil {
-			rc = io.NopCloser(reqBody)
-		}
-		req.Body = rc
-		cl, err := io.Copy(io.Discard, rc)
+		pr, pw := io.Pipe()
+		cl, err := io.Copy(io.Discard, io.TeeReader(reqBody, pw))
 		if err != nil {
 			return err
 		}
 		req.ContentLength = cl
+		req.Body = pr
 
 		r.setContentTypeHeader(req)
 		for k, v := range r.headers {
--- a/http.go
+++ b/http.go
@@ -257,6 +257,10 @@ func (rnr *httpRunner) Run(ctx context.Context, r *httpRequest) error {
 			return err
 		}
 		// reset Request.Body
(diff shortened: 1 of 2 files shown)


k2tzumi:
Oh. What I feared happened.
https://github.com/k1LoW/runn/actions/runs/3601501427/jobs/6067375050#step:7:35

I couldn't reproduce it in the last test, but I didn't think it would be at a time like this. 😅

atsushi-ishibashi:
How about using io.TeeReader?

atsushi-ishibashi:
how about this?
```
bw := new(bytes.Buffer)
req, err = http.NewRequestWithContext(ctx, r.method, u.String(), io.TeeReader(reqBody, bw))
if err != nil {
	return err
}
r.setContentTypeHeader(req)
for k, v := range r.headers {
	req.Header.Set(k, v)
}
rnr.operator.capturers.captureHTTPRequest(rnr.name, req)
...

req, err = http.NewRequestWithContext(ctx, r.method, u.String(), bw)
if err != nil {
	return err
}
r.setContentTypeHeader(req)
for k, v := range r.headers {
	req.Header.Set(k, v)
}
res, err = rnr.client.Do(req)
if err != nil {
	return err
}
```

k2tzumi:
@atsushi-ishibashi 

Thank you for the suggested modifications.

The following PR resolved the problem
https://github.com/k1LoW/runn/pull/345

I will keep this in mind for future reference.

PR #345, opened 2022-12-05
Fix capture/runbook.go reading out request body when multipart/form-data

k1LoW:
Fix #342


k1LoW:
@k2tzumi 
It seems that the problem was that capture/runbook.go had read through the request body. I hope you can try it in your environment!

k2tzumi (approved):
Error resolved in my environment.
Thank you for the correction.

PR #347, opened 2022-12-05
Fix duplicate drainBody

k1LoW:


yaml.go
```diff
@@ -87,11 +87,6 @@ func CreateHTTPStepMapSlice(key string, req *http.Request) (yaml.MapSlice, error
 			{Key: contentType, Value: f},
 		}
 	case strings.Contains(contentType, MediaTypeMultipartFormData):
-		save, req.Body, err = drainBody(req.Body)
```
k1LoW, line 90: already drainBody at https://github.com/k1LoW/runn/pull/347/files#diff-acfe93d173e77894feaede834d0927584abe0fc551dba248d191e16edaa63611L48

PR #390, opened 2023-01-27
Print step description on failure if available

nobuyo:
Currently, the step descriptions do not seem to be printed when a scenario fails.
I think this PR will be useful for identifying the step that caused the failure more quickly when the number of steps is large or the scenario has high complexity.

before: 
```console
$ runn run testdata/book/failure_with_step_desc.yml
Always failure scenario with step desc ... failed to run testdata/book/failure_with_step_desc.yml: test failed on 'Always failure scenario with step desc'.steps[0]: (false) is not true
false
└── false => false


1 scenario, 0 skipped, 1 failure
```

after:
```console
$ runn run testdata/book/failure_with_step_desc.yml 
Always failure scenario with step desc ... failed to run testdata/book/failure_with_step_desc.yml: test failed on 'Always failure scenario with step desc'.steps[0] 'this is description': (false) is not true
false
└── false => false


1 scenario, 0 skipped, 1 failure
```

### Special note

The formatting of output may need some tidying up 🤔

### DIFF
--- a/operator.go
+++ b/operator.go
@@ -831,7 +831,11 @@ func (o *operator) runInternal(ctx context.Context) (rerr error) {
 					}
 					o.Debugf(cyan("Run '%s' on %s\n"), testRunnerKey, o.stepName(i))
 					if err := s.testRunner.Run(ctx, s.testCond); err != nil {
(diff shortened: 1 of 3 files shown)


k1LoW:
@nobuyo Thank you for your contribution!

PR #401, opened 2023-02-04
Support expand vars to output path of dump

### LINKED ISSUE #398: Allow out attribute of dump section to be specified by variable
For example, the following statement

```yml
vars:
   path: "path/to"
step:
   dump:
     expr: "dump"
     out: {{ vars.path }}
```

If this is already possible, I may be mistaken.

k2tzumi:
close #398


k2tzumi:
I tried to add a usage example to `dump.yml`, but quit because it would be environment-dependent and the file would remain.

k1LoW (approved):
LGTM!

PR #403, opened 2023-02-04
[BREAKING] Bump up github.com/antonmedv/expr version

k1LoW:


bind_test.go
```diff
@@ -131,11 +131,6 @@ func TestBindRunnerRunError(t *testing.T) {
 				loopCountVarKey: "reverved",
 			},
 		},
-		{
-			map[string]string{
-				"eval error": "1 / 0",
```
k1LoW, line 136: `1 / 0 = +Inf`

No error in newer versions of expr package

eval_test.go
```diff
@@ -93,7 +105,7 @@ func TestValues(t *testing.T) {
 		{`res.body.data.key contains "xxxxxx"`, []string{"res.body.data.key", `"xxxxxx"`}},
 		{`res.headers["Content-Type"] == "application/json"`, []string{`res.headers["Content-Type"]`, `"application/json"`}},
 		{`current.rows[0]`, []string{`current.rows[0]`}},
-		{`body[0]["key"].data`, []string{`body[0]["key"].data`}},
+		{`body[0]["key"].data`, []string{`body[0].key.data`}},
```
k1LoW, line 108: In newer versions of the expr package, foo.key and foo["key"] are the same AST.

Issue #409, opened 2023-02-09
Cannot test if elements are contained in array when mapping object array responded from an external web server


IzumiSy:
# Preparation
test.yaml
```yaml
desc: object array mapping
vars:
  expected: [10, 20]
runners:
  req: http://localhost:8000
steps:
  a:
    req:
      /resp.json:
        get:
          body: null
  b:
    test: vars.expected in map(steps.a.res.body, {#.value})
```
resp.json
```json
[
  { "value": 10 },
  { "value": 20 },
  { "value": 30 }
]
```
run a mock server responding with the resp.json
```bash
$ python3 -m http.server
```

# Result
```bash
$ runn run test.yaml --debug
Run 'req' on 'object array mapping'.steps.a
-----START HTTP REQUEST-----
GET /resp.json HTTP/1.1
Host: localhost:8000


-----END HTTP REQUEST-----
-----START HTTP RESPONSE-----
HTTP/1.0 200 OK
Connection: close
Content-Length: 60
Content-Type: application/json
Date: Thu, 09 Feb 2023 00:41:33 GMT
Last-Modified: Wed, 08 Feb 2023 08:14:49 GMT
Server: SimpleHTTP/0.6 Python/3.10.8

[
  { "value": 10 },
  { "value": 20 },
  { "value": 30 }
]

-----END HTTP RESPONSE-----

Run 'test' on 'object array mapping'.steps.b
object array mapping ... failed to run test.yaml: test failed on 'object array mapping'.steps.b: (vars.expected in map(steps.a.res.body, {#.value})) is not true
vars.expected in map(steps.a.res.body, {#.value})
├── vars.expected => [10,20]
├── map(steps.a.res.body, ) => ?
├── steps.a.res.body => [{"value":10},{"value":20},{"value":30}]
└──  => ?


1 scenario, 0 skipped, 1 failure
```

**Expected**: it should succeed, but looks like a compiler cannot understand types when using map function.

I don't understand why mapping says `map(steps.a.res.body, ) => ?`. It just can extract values by keys from an object array.

The another thing I have observed is that it works as expected when I don't use response coming from web servers but put the same response in `vars` instead. I don't know why but it looks typed.

k1LoW:
@IzumiSy Thank you for your report!

k1LoW:
By the way, does the test pass if follows?

```yaml
desc: object array mapping
vars:
  expected: [10, 20]
runners:
  req: http://localhost:8000
steps:
  a:
    req:
[...]

PR #413, opened 2023-02-09
Implement intersect function

IzumiSy:
Background: https://github.com/k1LoW/runn/issues/409#issuecomment-1423742426

I have decided to return intersected array instead of boolean, because that would be more flexible. 

test.yaml
```yaml
desc: intersect
vars:
  v1: [1, 2, 3]
  v2: [2, 3, 4]
steps:
  a:
    test: 2 in intersect(vars.v1, vars.v2)
  b:
    test: 3 in intersect(vars.v1, vars.v2)
  c:
    test: 4 in intersect(vars.v1, vars.v2)
```
result
```
intersect ... failed to run intersect.yaml: test failed on 'intersect'.steps.c: (4 in intersect(vars.v1, vars.v2)) is not true
4 in intersect(vars.v1, vars.v2)
├── 4 => 4
├── intersect(vars.v1, vars.v2) => [2,3]
├── vars.v1 => [1,2,3]
└── vars.v2 => [2,3,4]


1 scenario, 0 skipped, 1 failure
```

### DIFF
--- a/option_test.go
+++ b/option_test.go
@@ -873,6 +873,8 @@ func TestSetupBuiltinFunctions(t *testing.T) {
 		{"bool"},
 		{"time"},
 		{"compare"},
+		{"diff"},
+		{"intersect"},
 		{"sprintf"},
 	}
 	opt := Func("sprintf", fmt.Sprintf)
--- a/builtin/intersect.go
+++ b/builtin/intersect.go
@@ -1,0 +1,7 @@
+package builtin
+
+import "github.com/juliangruber/go-intersect"
+
+func Intersect(x, y interface{}) interface{} {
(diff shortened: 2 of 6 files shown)


option_test.go
```diff
@@ -873,6 +873,8 @@ func TestSetupBuiltinFunctions(t *testing.T) {
 		{"bool"},
 		{"time"},
 		{"compare"},
+		{"diff"},
+		{"intersect"},
```
IzumiSy, line 877: Added `diff` as well because here lacks that.

k1LoW (approved):
@IzumiSy Thank you!!! LGTM!

PR #423, opened 2023-02-11
Append basename function

k2tzumi:
The following functions are now built-in
https://pkg.go.dev/path/filepath#Base


k1LoW (approved):
LGTM!

PR #478, opened 2023-04-03
Support `force:` section to force all steps to run.

k1LoW:
Add an option to run to the last step even if the intermediate step fails.

If even one step fails, it is considered a failure as a runbook.

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,6 +425,14 @@ Skip all `test:` sections
 skipTest: true
 ```
 
+### `force:`
+
+Force all steps to run.
+
+``` yaml
+force: true
+```
+
 ### `loop:`
 
 Loop setting for runbook.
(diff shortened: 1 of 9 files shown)


k1LoW:
Hi @k2tzumi.

What do you think of `force:` as a section name?

katsumi-kato-LITALICO (approved):
The implementation implication is that it is a forced execution option as the term implies, so there was no discrepancy as far as I could see in the code.

I'll try to reflect this in the actual scenario and see if there is any final discrepancy.

k2tzumi (approved):
The implementation implication is that it is a forced execution option as the term implies, so there was no discrepancy as far as I could see in the code.

I'll try to reflect this in the actual scenario and see if there is any final discrepancy.

PR #482, opened 2023-04-06
Add Cloud Spanner support without `xo/dburl`

BIwashi:
Thank you for the great tool!

Since https://github.com/k1LoW/runn/pull/481 depended on an external package called `xo/dburl`, this library would need to be modified to work properly, but this PR has implemented an implementation that does not depend on `xo/dburl`.
If the `xo/dburl` fixes are merged, this PR and fix will no longer be necessary.

reference:
- https://github.com/xo/dburl/pull/26


db.go
```diff
@@ -49,7 +50,14 @@ type DBResponse struct {
 }
 
 func newDBRunner(name, dsn string) (*dbRunner, error) {
-	db, err := dburl.Open(normalizeDSN(dsn))
+	var db *sql.DB
+	var err error
+	if strings.HasPrefix(dsn, "sp://") {
```
k1LoW, line 55: Can you support `spanner://` as well as https://github.com/xo/dburl?

k1LoW (commented):
@BIwashi Thank you for your contribution!!

I have one request, and I will merge it as soon as it is fixed.

db.go
BIwashi: @k1LoW 
I researched and found that `spanner://` is also supported. (`sp://` was the alias that had been set)
So I fixed like this. https://github.com/k1LoW/runn/pull/482/commits/c690fad5e24856df4430b708ab46984ba7c144f2
What do you think?

- [dburl/scheme.go at b1661a2a1fc75d11485c718e112f09ed4778010f · xo/dburl](https://github.com/xo/dburl/blob/b1661a2a1fc75d11485c718e112f09ed4778010f/scheme.go#L95)
- [Go Playground - The Go Programming Language](https://go.dev/play/p/DRI8beVGrWP)

k1LoW (approved):
Thank you!!

Issue #496, opened 2023-05-06
Request fails with JSON files containing escape sequences for line feed codes


k2tzumi:
If you try to send a request for a json file that contains a newline code in the json value, the json itself will be sent as a string.

* json files containing line feed codes
https://github.com/k2tzumi/runn/blob/5cbb9f68afb2bc3344ec5aec094755ddeaa4f3dd/testdata/newline.json

* runbook that calls the json file
https://github.com/k2tzumi/runn/pull/20/files#diff-529bf6e17bec3f0713aa61973fdc8ff6ad93c059f1ba1650a90e498e582ab934

* result
https://github.com/k2tzumi/runn/actions/runs/4901420664/jobs/8752680939#step:7:132

* debug
  json body is not json!  
  ```console
  $ go run cmd/runn/main.go run testdata/book/httpbin_include.yml --debug
  .. snip ..
  
  Run 'req' on 'testing include'.steps[1]
  -----START HTTP REQUEST-----
  POST /post?count=0 HTTP/1.1
  Host: httpbin.org
  Content-Type: application/json
  X-Test: default
  
  "{\"bar\":\"abc\\r\\ndef\",\"foo\":\"abc\\ndef\"}"
  -----END HTTP REQUEST-----
  -----START HTTP RESPONSE-----
  HTTP/2.0 200 OK
  Content-Length: 581
  Access-Control-Allow-Credentials: true
  Access-Control-Allow-Origin: *
  Content-Type: application/json
  Date: Sat, 06 May 2023 12:17:56 GMT
  Server: gunicorn/19.9.0
  
  {
    "args": {
      "count": "0"
    }, 
    "data": "\"{\\\"bar\\\":\\\"abc\\\\r\\\\ndef\\\",\\\"foo\\\":\\\"abc\\\\ndef\\\"}\"", 
    "files": {}, 
    "form": {}, 
    "headers": {
      "Accept-Encoding": "gzip", 
      "Content-Length": "50", 
      "Content-Type": "application/json", 
      "Host": "httpbin.org", 
      "User-Agent": "Go-http-client/2.0", 
      "X-Amzn-Trace-Id": "Root=1-64564571-525c7cd93ad52c044831da9c", 
      "X-Test": "default"
    }, 
    "json": "{\"bar\":\"abc\\r\\ndef\",\"foo\":\"abc\\ndef\"}", 
    "origin": "20.205.207.201", 
    "url": "https://httpbin.org/post?count=0"
  }
  
[...]

PR #497, opened 2023-05-07
Fix an issue that the JSON value in the request body becomes a string  when it contains newlines.

k1LoW:
Fix https://github.com/k1LoW/runn/issues/496


k2tzumi (approved):
LGTM!

PR #499, opened 2023-05-08
Support for JSON type columns

xande0812:
Thank you for the great tool.

In my project, I am using PostgreSQL's JSONB column. I would love to use this tool for testing, but it doesn't handle JSONB columns well. So I made some modifications on my own. What do you think?

I feel like there is some repetition in the code for supporting JSONB and sqlite's JSON support. Do you have any good ideas? Or do you think it's fine as it is?

### DIFF
--- a/db.go
+++ b/db.go
@@ -139,10 +139,10 @@ func (rnr *dbRunner) Run(ctx context.Context, q *dbQuery) error {
 					return err
 				}
 				for i, c := range columns {
+					t := strings.ToUpper(types[i].DatabaseTypeName())
 					switch v := vals[i].(type) {
 					case []byte:
 						s := string(v)
-						t := strings.ToUpper(types[i].DatabaseTypeName())
 						switch {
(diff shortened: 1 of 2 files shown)


db.go
```diff
@@ -157,13 +158,32 @@ func (rnr *dbRunner) Run(ctx context.Context, q *dbQuery) error {
 								return fmt.Errorf("invalid column: evaluated %s, but got %s(%v): %w", c, t, s, err)
 							}
 							row[c] = d
+						case t == "JSONB": // PostgreSQL JSONB
```
k1LoW, line 161: > I feel like there is some repetition in the code for supporting JSONB and sqlite's JSON support. Do you have any good ideas? Or do you think it's fine as it is?

I'm fine with the implementation as it is 👍

k1LoW (approved):
LGTM! Thank you!!!

k1LoW (approved):
LGTM! Thank you!!!

xande0812:
Thank you for merging!

PR #508, opened 2023-05-17
Support yaml:// scheme for vars

IzumiSy:
I would like to load YAML files instead of JSON in the case when we have bulk structured data.


k1LoW (approved):
@IzumiSy Great!!! Thank you!!

Issue #534, opened 2023-06-09
In loop, json is recognized as string


t-kitamura-axel:
When I try to test bellow runn contents,  in roop, json is recognized as string

- runn contents
```
desc: xxxx xxxxx xxxx
vars:
  xxxxx_req:
    - "json://../../json/req/200/normal1.json"
    - "json://../../json/req/200/normal2.json"
  xxxxx_res:
    - "do some magic"
    - "do some magic"
steps:
  normal:
    desc:  xxxx xxxx xxxx xxxx
    loop:
      count: len(vars.xxxxx_req)
    req:
      /xxx/xxx/xx/xxxxx:
        post:
          headers:
            accept: "{{ vars.header.accept }}"
            Content-Type: "{{ vars.header.content_type }}"
            Host: "{{ vars.header.host }}"
            X-API-KEY: "{{ vars.header.x_api_key }}"
          body:
            application/json: "{{ vars.xxxxx_req[i] }}"
    test: steps.normal.res.status == 200
      && steps.normal.res.body contains vars.xxxxx_res[i]
```

- log 

```
-----START HTTP REQUEST-----
POST xxx/xx/xxx/xx HTTP/1.1
Host: xxxxx
Accept: application/json
Content-Type: application/json
X-Api-Key: xxxxxxx

"json://../../json/req/200/normal2.json" <- not json but string  

-----END HTTP REQUEST-----
```

In no roop, json is recognized correctly.

- runn contents

```
desc: xxx xxxx xxxx
vars:
  xxx_req: "json://../../json/req/200/normal1.json"
 steps:
  normal:
    desc:  xxx xxx xxx
    req:
      /xxx/xxx/xxx/xx:
        post:
          headers:
            accept: "{{ vars.header.accept }}"
            Content-Type: "{{ vars.header.content_type }}"
            Host: "{{ vars.header.host }}"
            X-API-KEY: "{{ vars.header.x_api_key }}"
          body:
            application/json: "{{ vars.xxx_req }}"
    test: steps.normal.res.status == 200
      && steps.normal.res.body contains "do some magic"
```

- log

```
-----START HTTP REQUEST-----
POST xxx/xxx/xxx/xx HTTP/1.1
Host: xxxx
Accept: application/json
Content-Type: application/json
X-Api-Key: xxxxxx

{"session": ${ normal1.json contents } <- json is json
-----END HTTP REQUEST-----

```

k1LoW:
[...]

PR #545, opened 2023-06-12
Support wildcards in `json://` and `yaml://` so that values can be retrieved with slice

k1LoW:
ref: #534 

## Proposal

Support wildcards in `json://` and `yaml://` so that values can be retrieved with slice.

Only when a wildcard ( `*` ) is used, the expanded values become a slice.

### DIFF
--- a/vars_test.go
+++ b/vars_test.go
@@ -63,6 +63,13 @@ func TestEvaluateSchema(t *testing.T) {
 			"",
 			true,
 		},
+		{"json://testdata/vars*.json", nil, []any{
+			map[string]any{"foo": "test", "bar": float64(1)},
+			[]any{
+				map[string]any{"foo": "test1", "bar": float64(1)},
+				map[string]any{"foo": "test2", "bar": float64(2)},
+			},
+		}, false},
 	}
 	for i, tt := range tests {
 		tt := tt
--- a/vars.go
+++ b/vars.go
@@ -3,13 +3,18 @@ package runn
 import (
 	"bytes"
 	"fmt"
+	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 	"text/template"
 
+	"github.com/bmatcuk/doublestar/v4"
 	"github.com/goccy/go-json"
 	"github.com/goccy/go-yaml"
 )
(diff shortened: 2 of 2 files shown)


vars_test.go
```diff
@@ -63,6 +63,13 @@ func TestEvaluateSchema(t *testing.T) {
 			"",
 			true,
 		},
+		{"json://testdata/vars*.json", nil, []any{
```
k1LoW, line 66: vars.json and vars_array.json

k1LoW:
@k2tzumi If you don't mind, I'd like to hear your opinion.

k2tzumi:
If pattern matching with wildcards and . / (dot-dot-slash), there is a concern that directory traversal vulnerabilities may occur.

I have the impression that it would be difficult to use wild card matching because the order of appearance on the slice would be indefinite.

It seems to me that the goal of this PR should be more clearly defined.

* Solution to prevent vars from being fat
* Better Data Management Methods for Data Driven Testing

k1LoW:
Thank you for your advice!

> If pattern matching with wildcards and . / (dot-dot-slash), there is a concern that directory traversal vulnerabilities may occur.

Since we already allow dot-dot-slash, I thought it would be on the same level if wildcard was allowed?

> because the order of appearance on the slice would be indefinite.

https://github.com/k1LoW/runn/pull/545/files#diff-29e5d1896212c61371d2eb62885b72355eb956d2b92da55ae85b11bcaa5212cfR66

It sorts the matched files. I think that this correspondence will at least fix the order.

> It seems to me that the goal of this PR should be more clearly defined.

You are absolutely right.

Looking at #534,
I thought there was a request for one JSON file per `application/json HTTP request`, and one JSON file per `application/json HTTP response`.

I thought this PR might be a good idea since it is a feature enhancement and I can confirm that the modification will not affect existing implementations.

However, I have not thought deeply about this PR yet, so there may be other problems.

k1LoW:
Some of the security concerns can now be suppressed by [the scope](https://github.com/k1LoW/runn#scope).

Therefore, we are merging this feature.

PR #547, opened 2023-06-18
Enable configurable http runner timeout

### LINKED ISSUE #512: Enable configurable HttpRunner timeout
I would like to be able to configure from httpRunnerOption.
https://github.com/k1LoW/runn/blob/3fef50f2d647fad63f91cf0e4fc3f1a01a375b9a/http.go#L81

k2tzumi:
close #512


testdata/book/http_timeout.yml
```diff
@@ -0,0 +1,20 @@
+desc: Test using HTTP(Timeout)
+runners:
+  req:
+    endpoint: ${HTTPBIN_END_POINT:-https://httpbin.org/}
+    openapi3: https://tryapisproxy.com/spec/httpbin
+    timeout: 1
```
k2tzumi, line 6:
```suggestion
    timeout: 1s
```

Should the conversion from string be done with the time.ParseDuration function?
I think most of the use cases we specify are in seconds.

k1LoW: I would like to see you use [duration.Parse](https://github.com/k1LoW/duration) in the same way as in the `loop.interval:` section.
Because I want to align the way to set the time in the runbook.

option.go
```diff
@@ -196,6 +197,12 @@ func Runner(name, dsn string, opts ...httpRunnerOption) Option {
 			r.client.CheckRedirect = notFollowRedirectFn
 		}
 		r.multipartBoundary = c.MultipartBoundary
+		if c.Timeout != "" {
+			r.client.Timeout, err = duration.Parse(c.Timeout)
+			if err != nil {
+				return fmt.Errorf("timeout in HttpRunnerConfig is invalid: %w", err)
+			}
+		}
```
k2tzumi, line 205: What about putting in a timeout setting before calling the newHttpValidator function?
It seems like unnecessary logic or duplicate code, what do you think?

testdata/book/http_timeout.yml
```diff
@@ -0,0 +1,22 @@
...
+    test: |
+      current.res.status == 200
+  timout:
+    req:
+      # context deadline exceeded (Client.Timeout exceeded while awaiting headers)
+      /delay/3:
+        get:
+          body: null
+    test: |
+      current.res.status == 200
```
k1LoW, line 22: Looks like it needs a new line at the end.

option.go
k1LoW: > It seems like unnecessary logic or duplicate code, what do you think?

I think the duplication of code is an issue with the current runn structure, so there is no way around it.

k1LoW (approved):
LGTM!

testdata/book/http_timeout.yml
k1LoW: Fix confirmed 👍

k1LoW (approved):
Thank you!!

Issue #548, opened 2023-06-23
Allow HTTPRunner to take over cookies


k2tzumi:
Currently, you have to explicitly create a scenario to take over cookies, but it would be nice to have a mode to easily take over.

k1LoW:
I agree with your opinion 👍

k2tzumi:
After much discussion on the whiteboard, we came to an agreement on what we would do 👍 

![FzcmuOBaQAIRiM-](https://github.com/k1LoW/runn/assets/1182787/715fca99-f78d-4b80-bc39-d2330c2f967e)

k1LoW:
@k2tzumi GREAT WORK!!! Thank you!!

PR #549, opened 2023-06-23
Append Inherit cookies Example

k2tzumi:
refs #548

### DIFF
--- a/testdata/book/cookie.yml
+++ b/testdata/book/cookie.yml
@@ -1,0 +1,38 @@
+desc: Inherit cookies Example
+runners:
+  req:
+    endpoint: ${HTTPBIN_END_POINT:-https://httpbin.org/}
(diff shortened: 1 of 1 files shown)


testdata/book/cookie.yml
```diff
@@ -0,0 +1,38 @@
...
+  bindCookies:  
+    bind:
+      cookies: steps.setCookie.res.headers["Set-Cookie"]
+  inheritCookie:
+    desc: Returns cookie data.
+    req:
+      /cookies:
+        get:
+          headers:
+            Cookie: "{{ cookies[0] }};"
```
k2tzumi, line 34: The cookie header must be generated by interpreting the contents of the Set-Cookie and taking into account the case where a path or "expire" is specified.

k2tzumi:
The following PR responses are expected to make cookies a little simpler to handle
https://github.com/k1LoW/runn/pull/556

Issue #555, opened 2023-07-15
Define an accessor to get the Host of the endPoint of HTTPRunner


k2tzumi:
For example, it allows access to

```
httpRunners['req'].endpoint.Host
```

We would like to use it in a use case that identifies the domain of the cookie object

k1LoW:
I have one question. Is this functionality necessary because the endpoint is set by environment variables, etc. and is not fixed to one?

In most cases, the endpoint should be known by the user executing the runn, so I am cautious about the need for this additional functionality.

---

I think `httpRunners['req'].endpoint` could be generalized as simply "allow access to the Runner's settings".

I am wondering if we should have access to the hostname in `.Host` or not.

Perhaps it would be better to make `urlhostname()` or `url.Parse()` a builtin-function.

k1LoW:
> I think httpRunners['req'].endpoint could be generalized as simply "allow access to the Runner's settings".

I think httpRunners['req'].endpoint would be easier to understand if it were a `string` entered by the user instead of `*url.URL`.

k2tzumi:
> Perhaps it would be better to make urlhostname() or url.Parse() a builtin-function.

The `url` function was incorporated and implemented.
https://github.com/k1LoW/runn/pull/556#discussion_r1264607087

Some redundancy remained in the area of including default values for environment variables.

> I am wondering if we should have access to the hostname in .Host or not.

https://vscode.dev/github/k1LoW/runn/blob/vulnerability-check/http.go#L47

The current `httpRunner` was carrying it around as a url, so I was thinking of putting it directly into the `store`.

k1LoW:
> The url function was incorporated and implemented.
> https://github.com/k1LoW/runn/pull/556#discussion_r1264607087

👍 Thank you!

> Some redundancy remained in the area of including default values for environment variables.

I am wondering if this redundancy is acceptable.

> The current httpRunner was carrying it around as a url, so I was thinking of putting it directly into the store.

I think that is fine if we only consider the ENDPOINT.

However, if we were to add this functionality, I would want to raise the level of abstraction in the form of "access to Runner's settings".

At this point, it is debatable whether the parsed value should be set or not.

For example, we need to consider the following values.

- `endpoint` of HTTP Runner
- `openapi3` of HTTP Runner
- `key` of gRPC Runner
- `protos` of gRPC Runner

k2tzumi:
> I am wondering if this redundancy is acceptable.

[...]

PR #556, opened 2023-07-15
Keep cookies in store

k2tzumi:
refs #548


testdata/book/cookie.yml
```diff
@@ -0,0 +1,36 @@
...
+      current.res.status == 200
+      && current.res.cookies[0].Name == vars.cookie.key
+      && current.res.cookies[0].Value == uuid
+  inheritCookie:
+    desc: Returns cookie data.
+    req:
+      /cookies:
+        get:
+          headers:
+            Cookie: "{{ cookies['httpbin.org'][0].Raw }};"
```
k2tzumi, line 32: Hosts change from environment to environment, so we may need the ability to identify them.

#555

http.go
```diff
@@ -377,6 +378,21 @@ func (rnr *httpRunner) Run(ctx context.Context, r *httpRequest) error {
...
+	if len(cookies) > 0 {
+		keyMap := make(map[string]*http.Cookie)
+
+		for _, c := range cookies {
+			// If the Domain attribute is not specified, the host is taken over
+			c.Domain = rnr.endpoint.Host
+			keyMap[c.Name] = c
+		}
+
+		d[httpStoreCookieKey] = keyMap
```
k2tzumi, line 392: If there is no SetCookie in the response header, should I explicitly set an empty slice?

store_test.go
```diff
@@ -0,0 +1,249 @@
...
+					"example.com": {
+						// Override
+						"key1": &cookie4,
+					},
+				},
+			},
+			[]*http.Cookie{&cookie5},
+			map[string]map[string]*http.Cookie{
+				// Expire
+				"example.com": {},
```
k2tzumi, line 235: Once all cookies are deleted, should I also delete the domain key itself?

testdata/book/cookie.yml
```diff
@@ -0,0 +1,36 @@
...
+  setCookie:
+    desc: Returns a set of response headers from the query string.
+    req:
+      /response-headers?Set-Cookie={{ vars.cookie.key }}%3D{{ uuid }}:
+        get:
+          body: null
+    test: |
+      current.res.status == 200
+      && current.res.cookies[vars.cookie.key].Name == vars.cookie.key
+      && current.res.cookies[vars.cookie.key].Value == uuid
```
k2tzumi, line 25: When accessing the cookie of a step, the domain key is not needed because the domain of the request is obvious.

testdata/book/cookie.yml
```diff
@@ -0,0 +1,36 @@
...
+  inheritCookie:
+    desc: Returns cookie data.
+    req:
+      /cookies:
+        get:
+          headers:
+            Cookie: "{{ cookies[url('${HTTPBIN_END_POINT:-https://httpbin.org/}').Host][vars.cookie.key].Raw }};"
```
k2tzumi, line 32: I would like to consider addressing this in the following Issue as it is a redundant description
#555

http.go
k1LoW: I don't think empty slice is necessary because it doesn't become panic even if it is nil, but I thought it would be better to have empty map set.

https://github.com/k1LoW/runn/blob/867ca0cff63c8b5e7a2eb51e723664728cd18260/store.go#L131

https://github.com/k1LoW/runn/blob/867ca0cff63c8b5e7a2eb51e723664728cd18260/store.go#L155-L157

store.go
```diff
@@ -106,6 +110,28 @@ func (s *store) recordToLatest(key string, value any) error {
...
+		domain := cookie.Domain
+		if domain == "" {
+			domain = "localhost"
+		}
+		keyMap, ok := cookieMap[domain]
+		if !ok || keyMap == nil {
+			keyMap = make(map[string]*http.Cookie)
+		}
+		if !cookie.Expires.IsZero() && cookie.Expires.Before(time.Now()) {
+			// Remove expired cookie
```
k1LoW, line 125: Great!

store_test.go
k1LoW: I have no strong opinion on this point.

So I tried it in Chrome.

I erased all cookies, but the domain on the display remained.
CleanShot 2023-07-17 at 08 11 01@2x

testdata/book/cookie.yml
k1LoW: Great!!!

testdata/book/cookie.yml
```diff
@@ -0,0 +1,36 @@
...
+          body: null
+    test: |
+      current.res.status == 200
+      && current.res.body.cookies[vars.cookie.key] == uuid
```
k1LoW, line 36: Looks like it needs a new line at the end.

CleanShot 2023-07-17 at 08 13 27@2x

k1LoW (approved):
GREAT!!!

I have one modification request, but it is basically LGTM!

store_test.go
k2tzumi: Thank you for looking into this!
Gives me more confidence in these specs.

http.go
k2tzumi: Fixed this
https://github.com/k1LoW/runn/pull/556/commits/9d2c60c6104a55e1297a0011b9f2120a25ab83af

testdata/book/cookie.yml
k2tzumi: https://github.com/k1LoW/runn/pull/556/commits/c9c5f9e7b695187ffef946d55365f245b85c8cc2

k1LoW (approved):
LGTM!

k2tzumi:
I would like to consider including it in the README in phase 2
https://github.com/k1LoW/runn/issues/548#issuecomment-1606016518

PR #557, opened 2023-07-16
Fix `make lint` to run govulncheck locally as well

k1LoW:


Makefile
```diff
@@ -44,7 +45,7 @@ build:
 depsdev:
 	go install github.com/Songmu/ghch/cmd/ghch@latest
 	go install github.com/Songmu/gocredits/cmd/gocredits@latest
-	go install github.com/securego/gosec/v2/cmd/gosec@latest
+	go install golang.org/x/vuln/cmd/govulncheck@latest
```
k2tzumi, line 48: Since you no longer use gosec, do you no longer need to suppress the following warning?
https://vscode.dev/github/k1LoW/runn/blob/enable-configurable-http-runner-timeout/operator.go#L1356

k2tzumi (approved):
LGTM!
Thanks.

Makefile
k1LoW: gosec is actually running in golanci-lint 👍 

https://github.com/k1LoW/runn/blob/e7456e1e6ad594e630476ca80abdaf46844cbd86/.golangci.yml#L5

k2tzumi: nrhd-

k1LoW:
Oops 💦 
https://github.com/k1LoW/runn/actions/runs/5570516319/jobs/10174637026#step:7:87

PR #559, opened 2023-07-17
Append use cookie option

k2tzumi:
refs #548


http.go
```diff
@@ -308,6 +349,12 @@ func (rnr *httpRunner) Run(ctx context.Context, r *httpRequest) error {
 			return err
 		}
 		r.setContentTypeHeader(req)
+
+		// Override useCookie
+		if r.useCookie == nil && rnr.useCookie != nil && *rnr.useCookie {
+			r.useCookie = rnr.useCookie
+		}
```
k2tzumi, line 356: httpRunner setting is enabled only if the useCookie option is unset in the req section

http.go
```diff
@@ -55,6 +56,7 @@ type httpRunner struct {
 	cert              []byte
 	key               []byte
 	skipVerify        bool
+	useCookie         *bool
```
k2tzumi, line 59: The setting of httpRunner may not need to be nil, but httpRequest handles nil, so we have matched it.

testdata/book/cookie_in_requests_automatically.yml
```diff
@@ -0,0 +1,45 @@
+desc: Cookie in requests automatically
+runners:
+  req:
+    endpoint: ${HTTPBIN_END_POINT:-https://httpbin.org/}
+    useCookie: true
```
k1LoW, line 5: 👀

k1LoW (approved):
LGTM!

PR #566, opened 2023-07-21
fix typo

okazaki-kk:
typo maybe...?


k1LoW:
@okazaki-kk Thank you!!

PR #575, opened 2023-08-02
Fix a problem that functions with uint type arguments can't be used

n3xem:
* make the functions argument type int because int type can be used in runbook
* return `""` when less than 0 is passed to the argument
* add test because logics is added


k1LoW:
@n3xem Thank you!!!

k1LoW:
It looks like 1.20.7 won't be applied until https://github.com/actions/go-versions/pull/80 is merged.

PR #581, opened 2023-08-04
Improve error message when `runn run` fails

k1LoW:


k1LoW:
Cannot comment if not in the Pull Request changed file...

k1LoW:
The following is the current progress

CleanShot 2023-08-05 at 00 50 42@2x

k1LoW:
The following is the current progress

CleanShot 2023-08-05 at 11 18 17@2x

PR #587, opened 2023-08-08
Fix gRPC request message being pre-expanded values

k1LoW:


testdata/book/grpc.yml
```diff
@@ -9,7 +14,7 @@ steps:
         headers:
           authentication: tokenhello
         message:
-          name: alice
+          name: "{{ vars.names[0] }}"
```
k1LoW, line 17: Before the fix, variables were not being expanded and were showing up in the debug log.

PR #597, opened 2023-08-16
Use bufbuild/protocompile instead of jhump/protoreflect

k1LoW:


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/k1LoW/runn
 
-go 1.20
+go 1.21
+
+toolchain go1.21.0
```
k2tzumi, line 5: 👍

PR #600, opened 2023-08-17
Migrate to built-in functions of expr

k1LoW:
Deprecate some functions to use the built-in functions added at https://github.com/antonmedv/expr/releases/tag/v1.14.0

- `base64encode` -> `toBase64`
- `base64decode` -> `fromBase64`
- `json.Encode` -> `toJSON`
- `json.Decode` -> `fromJSON`


k1LoW:
@k2tzumi 
I would prefer to keep only the expr built-in functions, what do you think?

k2tzumi:
> I would prefer to keep only the expr built-in functions, what do you think?

I think it’s good.

Issue #606, opened 2023-08-31
Too many connections error occurs when multiple scenarios run


cxconj:
runn v0.81.1

Because a db connection is created each scenario and it is not closed after running the scenario, `Too many connections` error occurs when exceeding max connection limit.
Example: https://gist.github.com/cxconj/126d0848e62c273b06be037576f6cfd4

If the connection is closed after each scenario, the error will not occur, but it will affect not only the `runn` command but also the functions when runn is used as go pkg.
https://github.com/cxconj/runn/commit/327fa9fc393ddadf70ec255e61f82405a05d6c83

k1LoW:
@cxconj Thank you for your report!!

> but it will affect not only the runn command but also the functions when runn is used as go pkg.

You are right.
We'll try to find a solution.

k1LoW:
@cxconj 
Thanks to you, including providing the code for reference, I was able to solve the problem. Thank you very much.

cxconj:
Thanks for the quick fix! I have confirmed that it works correctly.

PR #607, opened 2023-08-31
Fix DB Runner connections are not closed properly

k1LoW:
ref: #606 

By this Pull Request, DB Runner with DSN disconnects from the database for each scenario.


k1LoW:
Reproduced too many connections error in test

https://github.com/k1LoW/runn/actions/runs/6037105698/job/16380717458?pr=607#step:8:54

k1LoW:
Re-reproduced too many connections error in test

https://github.com/k1LoW/runn/actions/runs/6037105698/job/16380717458?pr=607#step:8:54

PR #617, opened 2023-09-23
Support application/octet-stream via http

DaichiUeura:
This PR is to support "application/octet-stream" via http.

Here is the runbook sample:
```
    req:
      /upload:
        post:
          body:
            application/octet-stream:
              filename: ../dummy.jpg
```
This sample POSTs the (binary) file specified by ```filename``` with application/octet-stream content-type.


k1LoW:
@DaichiUeura Thank you for your excellent contribution!!

PR #640, opened 2023-10-13
Update README with interval option

IzumiSy:
`interval` options is not described in README so this PR adds it.


k1LoW:
@IzumiSy Oh! Thank you!

PR #645, opened 2023-10-15
Add header for trace

k2tzumi:
refs #619

Issue #619, opened 2023-09-25
Custom header for HttpRunner traceability


k2tzumi:
I would like to add a custom header to the request header to ensure traceability to the HTTP request during scenario execution.

We would like to be able to trace at least on a scenario-by-scenario basis from the application under test.

The specific goal is to set a unique string for each scenario in the request header when performing code coverage for API scenario tests, and to measure coverage based on that header information.

Considerations include the following

* How to specify the granting of custom headers?
* How do we support turning on/off the assignment of custom headers?
* How much flexibility do we allow for custom header strings?  
Do we want to automatically generate unique strings on a rule basis?

k1LoW:
Nice idea!

I think the most minimal and least impactful implementation would be to allow custom headers to be registered in the HTTP Runner configuration in the `runners:` section.

k2tzumi:
I would like to summarize the functional ideas for setting a custom header.
I think it would be good if parameters based on the attributes of the runbook being executed could be embedded, not just fixed strings. 
What parameters would be useful? 
I would appreciate your opinions on feasibility and performance.

* Path of the runbook
* Description of the runbook itself
* Name of the step being executed
* Description of the step being executed
* Method name and PATH specified for HttpRunner
* etc..

Also, is it okay to use antonmedv/expr as a library to evaluate these parameters? 
Is it okay to add keys to the store structure to access these parameters? 
Since there is a possibility that unintended headers may be embedded and cause security problems, should we manage the parameters in a separate structure?

k1LoW:
> I think it would be good if parameters based on the attributes of the runbook being executed could be embedded

This is too much for a feature.

It would be simpler overall to separate the features to be implemented.

1. Custom headers in the HTTP runner configuration
2. runners: section to accept dynamic values (This is very hard to do)
3. Provide variables to access runn and runnbook attributes

At this point, I feel there is not enough need for 2 and 3.

Because for 2, environment variables can dynamically change the value of the runners section. It also accepts variables only in the included runbook.
I feel that these features are enough for me.

Because for 3, I am cautious about adding a reserved variable because we have not gotten enough use cases.

How about we start with 1 first?

k2tzumi:
I agree with extending the functionality step by step.

To organize the steps, could you please confirm if the above steps are accurate?

> 1. Custom headers in the HTTP runner configuration

```yaml
desc: Test Custom headers 
runners:
  req:
    endpoint: ${HTTPBIN_END_POINT:-https://httpbin.org/}
[...]

PR #654, opened 2023-10-26
Add comment to query statement for trace

k1LoW:
ref: #619 #645 

SSIA

### DIFF
--- a/runner_option.go
+++ b/runner_option.go
@@ -171,7 +171,7 @@ func UseCookie(use bool) httpRunnerOption {
 	}
 }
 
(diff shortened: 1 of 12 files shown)


runner_option.go
```diff
@@ -171,7 +178,7 @@ func UseCookie(use bool) httpRunnerOption {
 	}
 }
 
-func Trace(trace bool) httpRunnerOption {
+func HTTPTrace(trace bool) httpRunnerOption {
```
k1LoW, line 181: This is breaking change

PR #663, opened 2023-10-30
Add the concept of scope.

k1LoW:
ref: #545 

And disallow remote reads and reads of directories above the working directory by default.


k1LoW:
Thank you @k2tzumi !

ref: https://x.com/katzchum/status/1719324320017854602?s=20

Issue #569, opened 2023-07-24
Access to response results during loop


k2tzumi:
For example.
```yaml
runners:
  req: https://api.example.com
steps:
  loopStep:
     loop:
      count: 10
    req:
      /odd?num={{ i }}
       get:
   referenceLoopStep:
     test: |
       steps.loopStep.loog[0].res.body.result == false  # I want to refer to the response result of each loop from another step.
       && steps.loopStep.loog[1].res.body.result == true
       && steps.loopStep.loog[2].res.body.result == false
```

k1LoW:
Maybe `s/loog/loop/` ?

k2tzumi:
yes. loop

k1LoW:
The namespace `loop` is not violated by anything else, and looks good to me.

On the other hand, if we can get a syntax that allows us to assign to an array with bind runner, we may be able to substitute it.

Then we would not have to increase namespace and memory usage would be limited.

```yaml
runners:
  req: https://api.example.com
steps:
  loopStep:
     loop:
      count: 10
    req:
      /odd?num={{ i }}
       get:
    bind:
      'loop[i]' : current.res # ?????????
   referenceLoopStep:
     test: |
       loop[0].res.body.result == false  # I want to refer to the response result of each loop from another step.
       && loop[1].res.body.result == true
       && loop[2].res.body.result == false
```

But I can't imagine a good syntax.

k2tzumi:
> if we can get a syntax that allows us to assign to an array with bind runner, we may be able to substitute it.

I thought the idea of being able to specify a variable with a variable name at the time of binding was very good. 👍

k2tzumi:
I thought the following description would be good

```yaml
runners:
  req: https://api.example.com
steps:
  loopStep:
     loop:
      count: 10
    req:
      /odd?num={{ i }}
       get:
    bind:
      "foo[]": current.res.body.foo 
```

k1LoW:
I tried to support two syntaxes, `foo[]` and `foo[i]` at #699.

k2tzumi:
Thank you!

PR #699, opened 2023-11-28
Support for bind to map values

k1LoW:
ref: https://github.com/k1LoW/runn/issues/569

```yaml
bind:
  vars:
    "foo[i]": current.res
```


k2tzumi (approved):
Now we can leave the result in the loop process!
Very good!

PR #711, opened 2023-12-23
Add workaround for go-sql-spanner internal connection sharing

### LINKED ISSUE #710: spanner: code = "InvalidArgument", desc = "invalid session pool"
Hi. I noticed an issue with the Spanner database integration of runn.

The runn implementation creates database client object for each runbooks to make the database connections are not shared between runbooks. However, the [go-sql-spanner driver reuses the same connection internally](https://github.com/googleapis/go-sql-spanner/blob/b56bfa8df549780852674f178a1c06f56f0f4c45/driver.go#L183) when the specified DSN is identical.

## Actual behavior
An error occurs when trying to use the same Spanner database in multiple runbooks.

```bash
❯ SPANNER_EMULATOR_HOST=localhost:9010 runn run  spanner_1.yml spanner_2.yml
..

1) spanner_2.yml 0e14c3d7ebcfe48aa13335c5d31fbd1681e0f4fb
  Failure/Error: db query failed on "Test using Spanner".steps[0]: spanner: code = "InvalidArgument", desc = "invalid session pool"
  Failure step (spanner_2.yml):
  5   - db:
  6       query: SELECT 1;


2 scenarios, 0 skipped, 1 failure
```

## Reproduction

1. Launch Spanner emulator
    ```bash
    docker run -p 9010:9010 -p 9020:9020 gcr.io/cloud-spanner-emulator/emulator
    ```
2. Setup the database
    ```bash
    docker run -eSPANNER_EMULATOR_HOST=localhost:9010 --network=host google/cloud-sdk:slim bash -c 'gcloud config configurations create emulator && gcloud config set auth/disable_credentials true && gcloud config set project test-project && gcloud config set api_endpoint_overrides/spanner http://localhost:9020/ && gcloud spanner instances create test-instance --config=emulator-config --description="Test Instance" --nodes=1 && gcloud spanner databases create test-database --instance=test-instance'
    ```
3. runn run
    ```bash
    SPANNER_EMULATOR_HOST=localhost:9010 runn run  spanner_1.yml spanner_2.yml
    ```

---

#### spanner_1.yml
```yaml
desc: Test using Spanner
runners:
  db: spanner://test-project/test-instance/test-database
steps:
  - db:
      query: SELECT 1;
```

#### spanner_2.yml
```yaml
desc: Test using Spanner
runners:
  db: spanner://test-project/test-instance/test-database
steps:
  - db:
      query: SELECT 1;
```

h6ah4i:
Fixes #710.

This workaround utilizes the [parameters part of the spanner DSN](https://github.com/googleapis/go-sql-spanner/blob/b56bfa8df549780852674f178a1c06f56f0f4c45/driver.go#L47) in order to generate different DSN string for each runbooks. The parameters part is properly handled in the [`newConnector()`](https://github.com/googleapis/go-sql-spanner/blob/b56bfa8df549780852674f178a1c06f56f0f4c45/driver.go#L195-L262) function, so it is safe to pass unsupported values like `workaroundConnectionInvalidationKey=123`.

### DIFF
--- a/db.go
+++ b/db.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"unsafe"
 
 	"github.com/araddon/dateparse"
@@ -14,7 +15,6 @@ import (
 	"github.com/golang-sql/sqlexp"
 	"github.com/golang-sql/sqlexp/nest"
 	_ "github.com/googleapis/go-sql-spanner"
-	"github.com/rs/xid"
 	"github.com/xo/dburl"
 	"modernc.org/sqlite"
 )
@@ -66,6 +66,7 @@ func newDBRunner(name, dsn string) (*dbRunner, error) {
 }
 
 var dsnRep = strings.NewReplacer("sqlite://", "moderncsqlite://", "sqlite3://", "moderncsqlite://", "sq://", "moderncsqlite://")
+var spannerInvalidatonKeyCounter uint64 = 0
 
 func normalizeDSN(dsn string) string {
 	if !contains(sql.Drivers(), "sqlite3") { // sqlite3 => github.com/mattn/go-sqlite3
@@ -287,9 +288,9 @@ func connectDB(dsn string) (TxQuerier, error) {
 	)
 	if strings.HasPrefix(dsn, "sp://") || strings.HasPrefix(dsn, "spanner://") {
 		// NOTE: go-sql-spanner trys to reuse the connection internally when the same DSN is specified.
-		key := xid.New().String()
+		key := atomic.AddUint64(&spannerInvalidatonKeyCounter, 1)
 		d := strings.Split(strings.Split(dsn, "://")[1], "/")
-		db, err = sql.Open("spanner", fmt.Sprintf(`projects/%s/instances/%s/databases/%s;workaroundConnectionInvalidationKey=%s`, d[0], d[1], d[2], key))
+		db, err = sql.Open("spanner", fmt.Sprintf(`projects/%s/instances/%s/databases/%s;workaroundConnectionInvalidationKey=%d`, d[0], d[1], d[2], key))
 	} else {
 		db, err = dburl.Open(normalizeDSN(dsn))
 	}
--- a/db.go
+++ b/db.go
@@ -3,7 +3,6 @@ package runn
 import (
 	"context"
 	"database/sql"
-	"encoding/base64"
 	"fmt"
 	"reflect"
 	"strconv"
@@ -15,6 +14,7 @@ import (
 	"github.com/golang-sql/sqlexp"
 	"github.com/golang-sql/sqlexp/nest"
 	_ "github.com/googleapis/go-sql-spanner"
+	"github.com/rs/xid"
 	"github.com/xo/dburl"
 	"modernc.org/sqlite"
 )
@@ -54,10 +54,14 @@ type DBResponse struct {
 }
 
 func newDBRunner(name, dsn string) (*dbRunner, error) {
+	nx, err := connectDB(dsn)
+	if err != nil {
+		return nil, err
+	}
 	return &dbRunner{
 		name:   name,
 		dsn:    dsn,
-		client: nil,
+		client: nx,
 	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


db.go
```diff
@@ -96,7 +93,7 @@ func (rnr *dbRunner) Run(ctx context.Context, s *step) error {
 func (rnr *dbRunner) run(ctx context.Context, q *dbQuery, s *step) error {
 	o := s.parent
 	if rnr.client == nil {
-		nx, err := connectDB(rnr.dsn)
+		nx, err := connectDB(rnr.dsn, s)
```
k1LoW, line 96: I think it's fine to generate them randomly without worrying about the steps.

db.go
```diff
@@ -53,14 +54,10 @@ type DBResponse struct {
 }
 
 func newDBRunner(name, dsn string) (*dbRunner, error) {
-	nx, err := connectDB(dsn)
```
k1LoW, line 56: Because newDBRunner() is executed for each runbook, how about generating a random DSN here for spanner only?

k1LoW (commented):
@h6ah4i GREAT! Thank you for your contribution!!!

I have a few suggestions and hope you'll consider them.

db.go
k1LoW: I think we don't have to worry about runbook IDs either.

h6ah4i: You are right. Using randomly generated numbers should work fine and I originally used them for this purpose. However, I changed my implementation because I thought using random numbers is not robust and it would be better to avoid if possible.

How do you think about the following code which is my initial workaround implementation.
```go
db, err = sql.Open("spanner", fmt.Sprintf(`projects/%s/instances/%s/databases/%s;workaroundConnectionInvalidationKey=%d`, d[0], d[1], d[2], rand.Int63()))
```

k1LoW: > because I thought using random numbers is not robust and it would be better to avoid if possible.

Indeed, I understand.

So how about [xid](https://github.com/rs/xid), which used to be used to generate IDs for runn?

https://github.com/k1LoW/runn/blob/4df90a92a72f4d684d86b79ad6f15773fe9dd1da/operator.go#L1161-L1163

It is more robust than random numbers because it uses time.

h6ah4i: Thanks, it seems xid is a good alternative! It has a **3-byte counter** part instead of the random number.

> 4-byte value representing the **seconds** since the Unix epoch,
> 3-byte machine identifier,
> 2-byte process id, and
> **3-byte counter**, starting with a random value

---

> It is more robust than random numbers because it uses time.

I think xid does offer better duplication resistance in theory, but not because using the time in this situation. The xid uses **second precision** unix epoch, however the `newDBRunner()` function is invoked for every runbooks at once during setup phase. It seems very large number of `newDBRunner()` function can be invoked within a second in contrast to the `func (rnr *dbRunner) run(ctx context.Context, q *dbQuery, s *step) error` function. The 3-byte counter part is more important in this situation.

h6ah4i:
@k1LoW I have updated my implementation to use xid for generating the invalidation key :tada:

db.go
k1LoW: Oops, second precision is not good.

I guess it would be better to use UUID or https://github.com/oklog/ulid or simply nano-precision time.

h6ah4i: I think xid is **good enough** in this situation. It has 3-bytes counter part, so it can be duplicated only if more than `2**24 = 16777216` xids generated per second.

```
1[sec] / (2^24)  =  1 [sec] / 16777216  ≒  60 [ns]
```

---

> use UUID or https://github.com/oklog/ulid 

I do not want to add more dependencies only for this workaround purpose...

---

> simply nano-precision time

I think this is difficult in practice. The timestamp resolution depends on the hardware and OS and [it does not have nano-precision time](https://stackoverflow.com/questions/57285292/why-does-time-now-unixnano-returns-the-same-result-after-an-io-operation) in most of time.

If you think 3 bytes counter is not enough, I recommend you to use a [64 bit atomic counter](https://pkg.go.dev/sync/atomic#Uint64.Add) instead of depending on timestamps.

k1LoW: > I do not want to add more dependencies only for this workaround purpose...

> I think this is difficult in practice. The timestamp resolution depends on the hardware and OS and [it does not have nano-precision time](https://stackoverflow.com/questions/57285292/why-does-time-now-unixnano-returns-the-same-result-after-an-io-operation) in most of time.

Thank you!
You are absolutely right. I am learning a lot.

> I recommend you to use a [64 bit atomic counter](https://pkg.go.dev/sync/atomic#Uint64.Add) instead of depending on timestamps.

GREAT!  I agree with this idea because it is simple.

h6ah4i: @k1LoW I have updated the code to use 64 bit atomic counter 😉

k1LoW:
Thank you for your patience and help with the revisions!!!!!

PR #713, opened 2023-12-29
Fix disable profile

k2tzumi:
refs https://github.com/k1LoW/runn/pull/677#pullrequestreview-1798968532

### DIFF
--- a/flags/flags.go
+++ b/flags/flags.go
@@ -81,6 +81,7 @@ func (f *Flags) ToOpts() ([]runn.Option, error) {
 		runn.GRPCNoTLS(f.GRPCNoTLS),
 		runn.GRPCProtos(f.GRPCProtos),
 		runn.GRPCImportPaths(f.GRPCImportPaths),
+		runn.DisableProfile(!f.Profile),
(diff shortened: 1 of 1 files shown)


flags/flags.go
```diff
@@ -81,6 +81,7 @@ func (f *Flags) ToOpts() ([]runn.Option, error) {
 		runn.GRPCNoTLS(f.GRPCNoTLS),
 		runn.GRPCProtos(f.GRPCProtos),
 		runn.GRPCImportPaths(f.GRPCImportPaths),
+		runn.DisableProfile(!f.Profile),
```
k2tzumi, line 84: Should the following line be true in addition?
https://github.com/k1LoW/runn/blob/4270638df6f7e338b5ef3224334b835bc412200a/cmd/run.go#L127

k1LoW: I want to enable profiles by default. So renaming `-profile` to `-disable-profie` is correct.

The reason for enabling profiles is to avoid missing information in result.json (Principle of least astonishment または Rule of least surprise) .

However, if it could have a significant impact on performance, it might be better to revert.

PR #716, opened 2023-12-30
Profile is default enabled and provides a way to disable it

k1LoW:
ref: https://github.com/k1LoW/runn/pull/713#discussion_r1438258885

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -76,7 +76,7 @@ var runCmd = &cobra.Command{
 			}
 		}
 
-		if flgs.Profile {
+		if !flgs.DisableProfile {
 			p, err := os.Create(filepath.Clean(flgs.ProfileOut))
 			if err != nil {
 				return err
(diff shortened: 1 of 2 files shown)


cmd/run.go
```diff
@@ -124,7 +124,7 @@ func init() {
 	runCmd.Flags().IntVarP(&flgs.ShardN, "shard-n", "", 0, flgs.Usage("ShardN"))
 	runCmd.Flags().IntVarP(&flgs.Random, "random", "", 0, flgs.Usage("Random"))
 	runCmd.Flags().StringVarP(&flgs.Format, "format", "", "", flgs.Usage("Format"))
-	runCmd.Flags().BoolVarP(&flgs.Profile, "profile", "", false, flgs.Usage("Profile"))
+	runCmd.Flags().BoolVarP(&flgs.DisableProfile, "disable-profile", "", false, flgs.Usage("DisableProfile"))
```
k2tzumi, line 127:
```suggestion
	runCmd.Flags().BoolVarP(&flgs.DisableProfile, "disable-profile", "", true, flgs.Usage("DisableProfile"))
```
What is the opposite of the opposite? Difficult.. 😅

k2tzumi: > Profile is default enabled

I was wrong.

k2tzumi (approved):
LGTM！

PR #718, opened 2023-12-30
Set up benchmark

k1LoW:
SSIA.

### DIFF
--- a/benchmark_test.go
+++ b/benchmark_test.go
@@ -1,0 +1,23 @@
+package runn
+
+import (
+	"context"
+	"testing"
+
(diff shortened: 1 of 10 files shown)


benchmark_test.go
```diff
@@ -0,0 +1,23 @@
+package runn
+
+import (
+	"context"
+	"testing"
+
+	"github.com/k1LoW/runn/testutil"
+)
+
+func BenchmarkRun(b *testing.B) {
```
k1LoW, line 10: @k2tzumi 
This is a sample benchmark.
Let's change it to a more valuable benchmark.

PR #721, opened 2023-12-30
Update expr (change org)

k1LoW:
Use https://github.com/expr-lang/expr instead of github.com/antonmedv/expr

CleanShot 2023-12-30 at 14 31 36@2x


k2tzumi (approved):
LGTM!

PR #724, opened 2023-12-30
Fix detectRunbookAreas() failure when the runbook has only 1 step in maps format

h6ah4i:
I noticed that `detectRunbookAreas()` function fails to parse the `steps` section in a specific situation.

When a runbook only has **single step with map-style**, the step will be parsed as `ast.MappingValueNode` which is not currently handled by runn.


k1LoW:
Thank you!!!!

PR #734, opened 2024-01-01
Update benchmarks

k1LoW:
Set benchmarks by number of runbooks, number of steps, and request/response body size.

### DIFF
--- a/testutil/runbook.go
+++ b/testutil/runbook.go
@@ -1,0 +1,72 @@
+package testutil
+
+import (
+	"fmt"
+	"html/template"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// BenchmarkSet returns a server and a path pattern for benchmarking.
+func BenchmarkSet(t testing.TB, bookCount, stepCount int, bodySize int) (*httptest.Server, string) {
+	const httpRunbookTmpl = `desc: Test using HTTP
+runners:
+  req: {{ .Endpoint }}
+steps:
+{{ range $k := .Steps }}
+-
+  req:
+    /:
+      post:
+        body:
+          application/json:
+            data: {{ $.Body }}
+{{ end }}
+`
+	ts := echoServer(t)
+	t.Cleanup(func() {
+		ts.Close()
+	})
+	dir := t.TempDir()
+	t.Cleanup(func() {
+		if err := os.RemoveAll(dir); err != nil {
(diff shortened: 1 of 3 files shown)


testutil/runbook.go
```diff
@@ -0,0 +1,71 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// BenchmarkSet returns a server and a path pattern for benchmarking.
+func BenchmarkSet(t testing.TB, bookCount, stepCount int, body string) (*httptest.Server, string) {
+	const httpRunbookTmpl = `desc: Test using HTTP
+runners:
+  req: {{ .Endpoint }}
```
k2tzumi, line 18: I want to enable the trace function and make it a bench target.

testutil/runbook.go
```diff
@@ -0,0 +1,71 @@
...
+steps:
+{{ range $k := .Steps }}
+-
```
k2tzumi, line 21: I want to set up a label and debug setting.

testutil/runbook.go
```diff
@@ -0,0 +1,71 @@
...
+  req:
+    /:
+      post:
+        body:
+          application/json:
+            {{ $.Body }}
+{{ end }}
```
k2tzumi, line 28: We want to increase coverage using as many different runners as possible.
test, bind, if, dump, eval, exec, loop etc..

k1LoW: The benchmarks for this are based on the number of runbooks, number of steps, and request/response body size = store size, and I believe that runner differences are out of scope.

testutil/runbook.go
k1LoW: Why do you want to enable `trace` ?

testutil/runbook.go
k1LoW: Why do you want to enable `label` or `debug`?

testutil/runbook.go
k2tzumi: It would be a good idea to set a policy for adding benchmarks.

* Make sure you have enough features to measure in one benchmark in advance.
Possibly detect unexpected performance degradation due to modifications
It is difficult to prepare benchmarks in advance. Hard to know which functions affected performance.
* Add a separate benchmark when you find a function for which you want to improve performance.
The more benchmarks you add, the longer it takes to run the CI
Which function's performance has changed? Easy to understand

My point was that I envisioned the former, and my understanding is that you were thinking the latter from the feel of our exchange on X.

PR #737, opened 2024-01-01
Change indentation options for flattening Yaml aliases

k2tzumi:
* before
image

* after
image

### DIFF
--- a/runbook.go
+++ b/runbook.go
@@ -131,7 +131,6 @@ func flattenYamlAliases(in []byte) ([]byte, error) {
 		goyaml.Flow(false),
 		goyaml.UseSingleQuote(false),
 		goyaml.UseLiteralStyleIfMultiline(false),
-		goyaml.Indent(1),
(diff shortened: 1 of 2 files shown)


runbook.go
```diff
@@ -131,7 +131,7 @@ func flattenYamlAliases(in []byte) ([]byte, error) {
 		goyaml.Flow(false),
 		goyaml.UseSingleQuote(false),
 		goyaml.UseLiteralStyleIfMultiline(false),
-		goyaml.IndentSequence(true),
+		goyaml.IndentSequence(false),
```
k1LoW, line 134: Am I correct in understanding that indentation is not necessary because it is in the middle of the process?

k2tzumi: Yes. We want to eliminate indentation during encoding and reduce the final byte size!

k1LoW:
LGTM!

k1LoW (approved):

PR #741, opened 2024-01-02
Reduce the number of execution of o.Result()

k1LoW:
SSIA.

### DIFF
--- a/operator.go
+++ b/operator.go
@@ -1425,16 +1425,15 @@ func (ops *operators) runN(ctx context.Context) (*runNResult, error) {
 			default:
 			}
 			defer func() {
+				r := o.Result()
+				o.capturers.captureResult(o.trails(), r)
+				o.capturers.captureEnd(o.trails(), o.bookPath, o.desc)
+				o.Close(false)
 				result.mu.Lock()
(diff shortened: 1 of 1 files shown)


operator.go
```diff
@@ -741,12 +741,12 @@ func (o *operator) Run(ctx context.Context) error {
 		o.sw.Disable()
 	}
 	defer o.sw.Start().Stop()
-	o.capturers.captureStart(o.trails(), o.bookPath, o.desc)
-	defer o.capturers.captureEnd(o.trails(), o.bookPath, o.desc)
-	defer o.Close(true)
 	defer func() {
 		o.capturers.captureResult(o.trails(), o.Result())
+		o.capturers.captureEnd(o.trails(), o.bookPath, o.desc)
+		o.Close(true)
```
k2tzumi, line 747: Is it because not including `o.sw.Start().Stop()` in an anonymous function changes the behavior? This question comes from my lack of understanding of the defer specification.

k1LoW: If we were to include it, the following

```go
o.sw.Start()
defer func() {
    o.sw.Stop()
}
```

Since I want to place Start() and Stop() in close proximity, I dare not include them in defer.

PR #768, opened 2024-01-20
Reflect skipValidateRequest when http-openapi3 option is enabled

k2tzumi:
refs https://zenn.dev/katzumi/scraps/f00a2d3e177b77#comment-36b36b5507b10d

### DIFF
--- a/operator.go
+++ b/operator.go
@@ -469,9 +469,17 @@ func New(opts ...Option) (*operator, error) {
 				if key != "" && key != k {
 					continue
 				}
-				val, err := newHttpValidator(&httpRunnerConfig{
+				ov, ok := bk.runners[k].(map[string]interface{})
+				if !ok {
+					return nil, fmt.Errorf("Invalid type: %v", bk.runners[k])
+				}
+				c := &httpRunnerConfig{
 					OpenApi3DocLocation: p,
-				})
+				}
+				c.SkipValidateRequest, _ = ov["skipValidateRequest"].(bool)
+				c.SkipValidateResponse, _ = ov["skipValidateResponse"].(bool)
+
+				val, err := newHttpValidator(c)
 				if err != nil {
 					return nil, err
 				}
--- a/operator.go
+++ b/operator.go
@@ -469,7 +469,7 @@ func New(opts ...Option) (*operator, error) {
 				if key != "" && key != k {
 					continue
 				}
-				runner, ok := bk.runners[k].(map[string]interface{})
+				runner, ok := bk.runners[k].(map[string]any)
 				if !ok {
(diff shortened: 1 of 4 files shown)


k2tzumi:
I tested it on a project with runn installed and it seemed to work fine.

operator.go
```diff
@@ -469,9 +469,17 @@ func New(opts ...Option) (*operator, error) {
 				if key != "" && key != k {
 					continue
 				}
-				val, err := newHttpValidator(&httpRunnerConfig{
+				runner, ok := bk.runners[k].(map[string]interface{})
```
k2tzumi, line 472: I don't think it will be any other type, but what do you think?

operator.go
```diff
@@ -469,9 +469,17 @@ func New(opts ...Option) (*operator, error) {
...
+				if !ok {
+					return nil, fmt.Errorf("invalid type: %v", bk.runners[k])
+				}
+				c := &httpRunnerConfig{
 					OpenApi3DocLocation: p,
-				})
+				}
+				c.SkipValidateRequest, _ = runner["skipValidateRequest"].(bool)
+				c.SkipValidateResponse, _ = runner["skipValidateResponse"].(bool)
```
k2tzumi, line 480: If it's not on yaml, it's assumed to be false.

k2tzumi (commented):
@k1LoW 
Is there anywhere else the parameters should be transferred?

operator.go
k1LoW, line 472: gostyle.ifacenames may be a false positive.

```suggestion
				runner, ok := bk.runners[k].(map[string]any)
```

operator.go
k1LoW: 👍

operator_test.go
```diff
@@ -1276,3 +1276,35 @@ func newRunNResult(t *testing.T, total int64, results []*RunResult) *runNResult
...
+	}{
+		{"false", true},
+		{"true", false},
+	}
+	for _, tt := range tests {
+		t.Run(fmt.Sprintf("skipValidateRequest: %s", tt.skipValidateRequest), func(t *testing.T) {
+			ctx := context.Background()
+			ts := testutil.HTTPServer(t)
+			t.Setenv("TEST_HTTP_END_POINT", ts.URL)
+			t.Setenv("SKIP", tt.skipValidateRequest)
```
k1LoW, line 1293:
```suggestion
			t.Setenv("TEST_SKIP_VALIDATE_REQUEST", tt.skipValidateRequest)
```

It is easy to see from the environment variable names which fields are reflected.

k1LoW (commented):
Thank you!!

I'm suggesting a very small correction.

k1LoW (approved):
LGTM!

PR #769, opened 2024-01-21
Keep loaded OpenAPI documents

k2tzumi:
refs https://zenn.dev/katzumi/scraps/f00a2d3e177b77#comment-24254b064a4eb5

### DIFF
--- a/http_validator.go
+++ b/http_validator.go
@@ -25,6 +25,12 @@ type UnsupportedError struct {
 	Cause error
 }
 
+var oasLoader *openapi3.Loader
+
+func init() {
+	oasLoader = openapi3.NewLoader()
+}
+
 func (e *UnsupportedError) Error() string {
 	return e.Cause.Error()
 }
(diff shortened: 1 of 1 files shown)


http_validator.go
```diff
@@ -25,6 +25,12 @@ type UnsupportedError struct {
 	Cause error
 }
 
+var oasLoader *openapi3.Loader
+
+func init() {
+	oasLoader = openapi3.NewLoader()
+}
```
k2tzumi, line 32: How can we implement singleton in golang?
I'm thinking that this process doesn't have to be that thread-safe.

k2tzumi (commented):
@k1LoW 
https://zenn.dev/katzumi/scraps/f00a2d3e177b77#comment-bc7c8680c8430d

Coverage commands also benefited

http_validator.go
```diff
@@ -63,31 +69,26 @@ type openApi3Validator struct {
...
 				return nil, err
 			}
 		default:
-			b, err := readFile(l)
-			if err != nil {
-				return nil, err
-			}
-			doc, err = loader.LoadFromData(b)
+			var err error
+			doc, err = oasLoader.LoadFromFile(l)
```
k2tzumi, line 85: `LoadFromData` was not cached.

k1LoW:
@k2tzumi 
GREAT WORK!!!!!!!!!!!!!! 🚀 🚀 🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

I am going to review it.

http_validator.go
k1LoW: NewLoader() may be executed only when openapi3.Loader is needed (e.g. sync.Once).

k1LoW: However, the performance is good enough to be better than it is now, so I think it is fine as it is.
(We can improve it another time.)

k1LoW (approved):
GREAT!

Issue #803, opened 2024-02-18
[Feature request] Add diff output when compare() returns false


h6ah4i:
Hi. I think it would be great if the built-in `compare()`  function prints human-friendly diff output when the evaluated result is `false`.

### Current output
```
❯ runn run ./example.yaml
F

1) ./example.yaml 660d883553d7b12515432bc36b550c485d82d3fd
  Failure/Error: test failed on "[No Description]".steps.useCompare "Use compare()": condition is not true

  Condition:
    compare(vars.element1, vars.element2)
    │
    ├── compare(vars.element1, vars.element2) => false
    ├── vars.element1 => {"bar":2,"baz":3,"foo":1}
    └── vars.element2 => {"bar":2,"foo":10}

  Failure step (./example.yaml):
   8   useCompare:
   9     desc: 'Use compare()'
  10     test: |
  11       compare(vars.element1, vars.element2)
  12

2) ./example.yaml 660d883553d7b12515432bc36b550c485d82d3fd
  Failure/Error: test failed on "[No Description]".steps.useDiff "Use diff()": condition is not true

  Condition:
    diff(vars.element1, vars.element2) == ''
    │
    ├── diff(vars.element1, vars.element2) => "  map[string]any{
    │     	"bar": float64(2),
    │   - 	"baz": float64(3),
    │   - 	"foo": float64(1),
    │   + 	"foo": float64(10),
    │     }
    │   "
    ├── vars.element1 => {"bar":2,"baz":3,"foo":1}
    ├── vars.element2 => {"bar":2,"foo":10}
    └── "" => ""

  Failure step (./example.yaml):
  13   useDiff:
  14     desc: 'Use diff()'
  15     test: |
  16       diff(vars.element1, vars.element2) == ''
  17


1 scenario, 0 skipped, 1 failure
```

### Proposed output (1)
```diff
❯ runn run ./example.yaml
F

1) ./example.yaml 660d883553d7b12515432bc36b550c485d82d3fd
  Failure/Error: test failed on "[No Description]".steps.useCompare "Use compare()": condition is not true

  Condition:
    compare(vars.element1, vars.element2)
    │
    ├── compare(vars.element1, vars.element2) => false
+    |   "  map[string]any{
+    │     	"bar": float64(2),
+    │   - 	"baz": float64(3),
+    │   - 	"foo": float64(1),
+    │   + 	"foo": float64(10),
+    │     }
+    │   "
[...]

PR #917, opened 2024-05-12
Replace the buildTree() method by injected expr tracing approach

h6ah4i:
Hi! I hope it will be a big step forward for runn and I'm so happy to introduce this improvement :tada:

### What's new?

- Offers better pretty printing output in tree format for expr evaluation results.
- The new implementation is based on **tracing** approach. All evaluation results including intermediate states of the expr programs are preserved in memory, so each expr program is not be evaluated again during the pretty printing phase.
- The tracing feature is tested by the [official expr testdata](https://github.com/expr-lang/expr/blob/master/testdata/examples.txt)

### Drawbacks

- Some performance overhead added due to tracing
- (breaking change) Removed YAML style `#` line comment handling. The behavior is changed in order to accept every valid syntax expr codes (the `#` symbol is a pointer/special variable marker in expr lang)

### Behind the scenes

At first, I was trying to implement my proposal of the issue #803. However, soon I realized the current `buildTree()` method is not robust and there is no enough room for future development. That is why I decided to re-implement the pretty printing feature for expr evaluation results in completely new way.


k1LoW:
@h6ah4i 
GREAT WORK!!!!!!!!! Thank you!!

First of all, thank you. I will now check the code.

> However, soon I realized the current buildTree() method is not robust and there is no enough room for future development. 

I agree with you.

> (breaking change) Removed YAML style # line comment handling.

This was granted at a time when expr-lang did not have the ability to comment, and should go away.
However, we cannot suddenly lose it, so I will try to be able to give a message that it is deprecated ( and replace `^#` to `//` 
 simply).

k1LoW (approved):
GREAT!

PR #925, opened 2024-05-19
Print diff info for the compare() built-in function (issue #803)

h6ah4i:
This pull request implements the proposal outlined in issue #803. When the `compare()` built-in function returns `false`, a diff output will automatically be added for convenience.


k1LoW (approved):
@h6ah4i GREAT WORK!!! Thank you!

Issue #941, opened 2024-05-28
[Feature Request] Allow passing enviroment variables by file


okazaki-kk:
Hello! 
When testing with runn in CLI mode in multi environments, it's tedious to switch some environment variables for each environment like this.

```
ENV1=STAGING_ENV1 ENV2=STAGING_ENV2 ENV3=STAGING_ENV3 runn run test.yaml.   # staging
ENV1=RPOD_ENV1    ENV2=PROD_ENV2    ENV3=PROD_ENV3    runn run test.yaml    # production
```

So, I want to manage environment variables by file and pass them.

my image
```
runn run test.yaml -e staging.yaml     #  staging
runn run test.yaml -e prod.yaml        # production
```

If possible, I think It would be useful if environment variables could be specified in a yaml file or .env file.
If you have any other ideas, please let me know 🙏 .

ref: https://github.com/postmanlabs/newman?tab=readme-ov-file#newman-run-collection-file-source-options -e option

k1LoW:
If it were to be implemented, it would be better to add the same `--env-file` option as the docker command.

PR #945, opened 2024-05-30
Support `--env-file` option for loading environment variables from a file.

k1LoW:
Fix https://github.com/k1LoW/runn/issues/941

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -143,7 +143,9 @@ func init() {
 	runCmd.Flags().BoolVarP(&flgs.RetainCacheDir, "retain-cache-dir", "", false, flgs.Usage("RetainCacheDir"))
 	runCmd.Flags().StringVarP(&flgs.WaitTimeout, "wait-timeout", "", "10sec", flgs.Usage("WaitTimeout"))
 	runCmd.Flags().StringVarP(&flgs.EnvFile, "env-file", "", "", flgs.Usage("EnvFile"))
-	runCmd.MarkFlagFilename("env-file")
+	if err := runCmd.MarkFlagFilename("env-file"); err != nil {
+		panic(err)
+	}
 	runCmd.Flags().BoolVarP(&flgs.Verbose, "verbose", "", false, flgs.Usage("Verbose"))
 	runCmd.Flags().BoolVarP(&flgs.Attach, "attach", "", false, flgs.Usage("Attach"))
 }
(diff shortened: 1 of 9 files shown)


k1LoW:
@k2tzumi 
It would be nice to see about this fix, including whether it is `--env-file` or not!

cmd/run.go
```diff
@@ -142,6 +142,10 @@ func init() {
 	runCmd.Flags().StringVarP(&flgs.CacheDir, "cache-dir", "", "", flgs.Usage("CacheDir"))
 	runCmd.Flags().BoolVarP(&flgs.RetainCacheDir, "retain-cache-dir", "", false, flgs.Usage("RetainCacheDir"))
 	runCmd.Flags().StringVarP(&flgs.WaitTimeout, "wait-timeout", "", "10sec", flgs.Usage("WaitTimeout"))
+	runCmd.Flags().StringVarP(&flgs.EnvFile, "env-file", "", "", flgs.Usage("EnvFile"))
```
k2tzumi, line 145:
```suggestion
	runCmd.Flags().StringVarP(&flgs.EnvFile, "env-file", "", ".env", flgs.Usage("EnvFile"))
```

The option name `env-file` reminded me of docker-compose and the default value seems to be `.env`!

k1LoW: If there is a default value, it will always try to read that file due to the behaviour of Flags.
This is not the intended behaviour.

k2tzumi: https://docs.docker.com/compose/environment-variables/set-environment-variables/#substitute-with-an-env-file

> The .env file is the default method for setting environment variables in your containers. 

If you do not want the above to be the standard behavior, that is OK.

k2tzumi (approved):

k1LoW:
@k2tzumi Thank you!

cmd/run.go
k1LoW: nrhd...

I don't want to default load .env.

Because I don't think the .env file in the project root is for runn.

Thank you!!!

k1LoW: There is room for debate, but I will keep it as it is for now.

k2tzumi: > Because I don't think the .env file in the project root is for runn.

Indeed, the .env in the repository root is for the application under test by runn.

You may use the same env content, or you may not want to use it.

PR #951, opened 2024-06-09
Add `--force-color` option for forcing colorized output

na3shkw:
Added an option to enable colorized output in non-tty output streams like GitHub Actions. When specifying this option, bypass the check for non-tty output streams by the [fatih/color](https://github.com/fatih/color) package.

Output without option:
```bash
$ runn run testdata/book/runn_0_success.yml | less
.

1 scenario, 0 skipped, 0 failures
```

Output with `--force-color` option:
```bash
$ runn run testdata/book/runn_0_success.yml --force-color | less
[32m.[0m

[32m1 scenario, 0 skipped, 0 failures
[0m
```

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -29,6 +29,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/fatih/color"
 	"github.com/k1LoW/donegroup"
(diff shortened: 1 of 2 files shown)


k1LoW:
@na3shkw Thank you!! NICE!!

Issue #1047, opened 2024-09-25
Ignore fields when compare test


hungpk-bap:
I'm using a YAML format to define test cases and running a compare command in Go like this:

```yaml
compare(current.res.message, vars.response, vars.ignoreField)
```
I’ve tried defining ignoreField in two different ways:

As a string:

```yaml
vars:
  ignoreField: "field1, field2"
```
As a list:

```yaml
vars:
  ignoreField:
    - field1
    - field2
```
Results:

In case 1, Go interprets the entire string ```("field1, field2")``` as a single parameter, which is not the desired behavior.
In case 2, I get an error: cannot use []interface{} as type string.
The compare function works fine if I declare ignoreField as a single value. However, when I try to ignore multiple fields, it doesn't work.

**My Proposal:**
I’m thinking about modifying the ```compare``` function to accept ```[]string``` instead of ```...string``` (variadic string parameters) to handle a list of fields.

Could you please provide feedback on this approach or suggest a better solution?

hankphan-dev:
I'm also having this problem when trying to ignore multiple fields in the response when comparing

hankphan-dev:
Looks seem you want use 
```yml
compare(current.res.message, vars.response, vars.ignoreField) 
```
for multiple scenarios?
Can you try applying a comparison command to each scenario? Because you have to pass a list of arguments to that command instead of passing an argument vars.ignoreField containing the values of the fields separated by commas
```yml
Scenario 1: (ignore 2 field)
compare(current.res.message, vars.response, vars.ignoreField[0], vars.ignoreField[1]) 

Scenario 2: (ignore 3 field)
compare(current.res.message, vars.response, vars.ignoreField[0], vars.ignoreField[1], vars.ignoreField[2]) 
```
It works with me

k1LoW:
Hi @hungpk-bap ( @hankphan-dev ). Thank you for your feedback.

In expr-lang, it certainly seems like `[]string` is better.
However, even if we update the function signature, it seems like we'll need to do it step by step.

hungpk-bap:
> Hi @hungpk-bap ( @hankphan-dev ). Thank you for your feedback.
> 
> In expr-lang, it certainly seems like `[]string` is better. However, even if we update the function signature, it seems like we'll need to do it step by step.

@k1LoW thank you for your feedback.
[...]

PR #1052, opened 2024-10-02
Update signature of compare/diff built-in functions step by step

k1LoW:
ref: https://github.com/k1LoW/runn/issues/1047

Because variadic arguments like Go cannot be used in expr-lang.

While accepting both arguments, deprecate the variadic argument type.

### DIFF
--- a/builtin/diff.go
+++ b/builtin/diff.go
@@ -26,6 +26,10 @@ func Diff(x, y any, ignores ...any) (string, error) {
 				}
 				ignoreSpecifiers = append(ignoreSpecifiers, s)
 			}
+		case []string:
+			for _, s := range v {
+				ignoreSpecifiers = append(ignoreSpecifiers, s)
+			}
 		default:
 			return "", fmt.Errorf("invalid ignore specifiers: %v", i)
 		}
--- a/builtin/diff.go
+++ b/builtin/diff.go
@@ -10,13 +10,18 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/itchyny/gojq"
+	"github.com/k1LoW/runn/internal/deprecation"
 )
 
 func Diff(x, y any, ignores ...any) (string, error) {
 	var ignoreSpecifiers []string
+	if len(ignores) > 1 {
+		deprecation.AddWarning("diff/compare", "diff/compare(x, y, ignores...string) is deprecated. Use diff/compare(x, y, ignores []string) instead.")
+	}
 	for _, i := range ignores {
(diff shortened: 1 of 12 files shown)


k1LoW:
@k2tzumi
I'd be glad to get your feedback on this change.

builtin/diff.go
```diff
@@ -10,9 +10,36 @@ import (
...
 	"github.com/itchyny/gojq"
+	"github.com/k1LoW/runn/internal/deprecation"
 )
 
-func Diff(x, y any, ignorePaths ...string) (string, error) {
+func Diff(x, y any, ignores ...any) (string, error) {
+	var ignoreSpecifiers []string
+	if len(ignores) > 1 {
+		deprecation.AddWarning("diff/compare", "diff/compare(x, y, ignores...string) is deprecated. Use diff/compare(x, y, ignores []string) instead.")
+	}
```
k2tzumi, line 20: I understood that it would be phased out.

builtin/diff.go
```diff
@@ -10,9 +10,36 @@ import (
...
+			deprecation.AddWarning("diff/compare", "diff/compare(x, y, ignores...string) is deprecated. Use diff/compare(x, y, ignores []string) instead.")
+			ignoreSpecifiers = append(ignoreSpecifiers, v)
+		case []any:
+			for _, vv := range v {
+				s, ok := vv.(string)
+				if !ok {
+					return "", fmt.Errorf("invalid ignore specifiers: %v", vv)
+				}
+				ignoreSpecifiers = append(ignoreSpecifiers, s)
+			}
```
k2tzumi, line 33: Currently accepting ANY for compatibility, so it needs to be checked.

builtin/diff.go
```diff
@@ -25,7 +52,7 @@ func Diff(x, y any, ignorePaths ...string) (string, error) {
 		return "", err
 	}
 
-	jqIgnorePaths, cmpIgnoreKeys := impl.splitIgnoreSpecifiers(ignorePaths)
+	jqIgnorePaths, cmpIgnoreKeys := impl.splitIgnoreSpecifiers(ignoreSpecifiers)
```
k2tzumi, line 55: This process will also be unnecessary in the future.

testdata/book/builtin_compare.yml
```diff
@@ -0,0 +1,16 @@
+desc: For compare() built-in function
+steps:
+  merge:
+    test: |
+      compare(
+        {"a": 1, "b": 3, "c": 5},
+        {"a": 1, "b": 2, "c": 4},
+        "b", "c"
+      )
```
k2tzumi, line 9: This test here will fail in the future.

PR #1199, opened 2025-03-28
fix: line not found

zxckaneki666:
# Problem
In some cases, the parser registers more rows for the test than it actually occupies. Because of this, if this test is the last one in the file, runn crashes completely without displaying the test results.
This problem is most serious when splitting tests into different files (include), because one file causes all subsequent tests to crash.
# What changed
Instead of outputting an error about “line not found”, assign the end of the test to the end of the file.


k1LoW:
@zxckaneki666 
Thank you!

PR #1270, opened 2025-07-14
chore(dockerfile): [Proposal] Add `curl` in Dockerfile image

t-miyak:
## Changes

- Install `curl` in Dockerfile

## Background

I would like to install curl in the Docker image for cases where I need to execute HTTP requests that are not supported by runn's functionality.

In my use case, I wanted to upload image data to a Cloud Storage Signed URL in the format like the below, but I couldn't figure out how to upload it as binary data.

```yaml
runners:
  gcs:
    endpoint: "https://storage.googleapis.com"

steps:
  upload_image:
    gcs:
      "{{ trimPrefix(signed_url, 'https://storage.googleapis.com') }}":
        put:
          headers:
            Content-Type: "image/jpeg"
          body:
            "application/octet-stream": "{{ image_data }}"  # <- I wanted to set binary data or read it from image file path.
```

I understand that I could install curl on top of the existing distributed container image, but it would be convenient if it were installed by default. I would appreciate it if this could be accepted if there is no problem.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,7 @@ RUN apt-get update \
     && apt-get install -y fonts-noto-cjk \
     && apt-get install -y chromium \
     && apt-get install -y git \
+    && apt-get install -y curl \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 


k1LoW:
@t-miyak Thank you for your proposal!

I have a question.
What would be the command if I were to upload using curl?

t-miyak:
@k1LoW 

Thank you for your response!

The curl command would be:

```bash
curl -X PUT \
  -H "Content-Type: image/jpeg" \
  --upload-file "/path/to/image.jpg" \
  "{{ signed_url }}"
```

k1LoW:
@t-miyak 

Sorry for the delay in response.
We have made it possible to upload images with runn.
Please check it out if you like.

https://github.com/k1LoW/runn/blob/main/testdata/book/http_image.yml

t-miyak:
@k1LoW 

Thank you for the improvement! It worked in my local scenario test 😄 

I think this PR is unnecessary, so I will close it.

PR #1271, opened 2025-07-16
Add support for enabling `--no-sandbox` through an environment variable

windyakin:
To avoid requiring privileged mode when running Chrome (Chromium) in a container, this change allows the `--no-sandbox` flag to be set via ~~`RUNN_DISABLE_CHROMIUM_SANDBOX`~~ `RUNN_DISABLE_CHROME_SANDBOX` environment variable.

### DIFF
--- a/cdp.go
+++ b/cdp.go
@@ -60,7 +60,7 @@ func newCDPRunner(name, remote string) (*cdpRunner, error) {
 		)
 	}
 
-	if os.Getenv("RUNN_DISABLE_CHROMIUM_SANDBOX") != "" {
+	if os.Getenv("RUNN_DISABLE_CHROME_SANDBOX") != "" {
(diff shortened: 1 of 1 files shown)


cdp.go
```diff
@@ -60,6 +60,10 @@ func newCDPRunner(name, remote string) (*cdpRunner, error) {
 		)
 	}
 
+	if os.Getenv("RUNN_DISABLE_CHROMIUM_SANDBOX") != "" {
```
k1LoW:
```suggestion
	if os.Getenv("RUNN_DISABLE_CHROME_SANDBOX") != "" {
```

How about setting the env name to `RUNN_DISABLE_CHROME_SANDBOX`?
CHROMIUM may be correct, but I would like to prioritize clarity.

windyakin: OK.

k1LoW:
@windyakin Thank you!!

PR #1284, opened 2025-08-12
Support passing Chrome launch options from book

windyakin:
Previously, it was not possible to pass launch flags to Chrome (CDP) from book.

This change adds support for specifying and passing these flags, enabling more flexible Chrome configurations.

example:

```yaml
runners:
  cc:
    # optional (default: chrome://new)
    addr: chrome://new 
    # new!
    flags:
      proxy-server: 127.0.0.1:8080
      proxy-bypass-list: <local>,*.google.com
```

I’ve added some basic tests (cdp_options_test.go) that seem to cover the change, but please let me know if more are needed.

### DIFF
--- a/book.go
+++ b/book.go
@@ -276,6 +276,14 @@ func (bk *book) parseRunner(k string, v any) error {
 			}
 		}
 
+		// CDP Runner
+		if !detect {
+			detect, err = bk.parseCDPRunnerWithDetailed(k, tmp)
+			if err != nil {
+				return err
+			}
+		}
+
 		// Include Runner
 		if !detect {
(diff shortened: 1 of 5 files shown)


k1LoW:
@windyakin Thank you for your contribution! 

Not that it has anything to do with your fix, but it looks like we'll have to wait for a response to https://pkg.go.dev/vuln/GO-2025-3829

k1LoW:
Could you update with rebase? Perhaps the vuln issue has been resolved.

CleanShot 2025-08-15 at 10  16 06@2x

k1LoW:
@windyakin GREAT!!! Thank you!!!

PR #1294, opened 2025-09-05
feat: add warning for loop.until + test combination

tokuhirom:
- Add runtime warning when both loop.until and test are used in the same step
- Add documentation in README explaining the execution order and the pitfall

This helps users avoid the common mistake of using test conditions with retry loops, where the test runs on each iteration and may cause premature failure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -711,6 +711,48 @@ steps:
 [...]
 ```
 
+#### Execution order within a loop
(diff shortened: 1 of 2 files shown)


k1LoW:
@tokuhirom Thank you for your contribution!!

Issue #1323, opened 2025-10-12
Add preserveTrailingSlash option for HTTP requests


helloworld753315:
### Summary

The HTTP runner currently removes trailing slashes (`/`) from endpoint paths automatically. This becomes problematic when some APIs strictly distinguish between paths with and without trailing slashes. Adding a step-level `preserveTrailingSlash` option would allow preserving trailing slashes when needed.

### Current Behavior

When executing HTTP requests in runn, trailing slashes at the end of endpoint paths are automatically removed, even when explicitly specified.

This occurs because the `mergeURL` function in `http.go` uses `path.Join()` ([source code](https://github.com/k1LoW/runn/blob/main/http.go#L595)), which internally calls `path.Clean()` ([documentation](https://pkg.go.dev/path#Join): "The result is Cleaned", [implementation](https://github.com/golang/go/blob/master/src/path/path.go#L172)). According to the official Go documentation, `path.Clean()` specifies that **"the returned path ends in a slash only if it is the root '/'"** ([reference](https://pkg.go.dev/path#Clean)), so trailing slashes are removed from non-root paths.

### Expected Behavior

The HTTP runner should:

1. Recognize when a trailing slash is specified in the path
2. Preserve the trailing slash when `preserveTrailingSlash` option is set to `true`
3. Maintain current behavior (remove trailing slashes) by default when the option is not specified or set to `false` (backward compatibility)

### Use Cases

This feature is essential for:

- APIs that strictly distinguish between paths with and without trailing slashes
- APIs that require trailing slashes (e.g., returning 404 when the trailing slash is missing)

### Problem Analysis

1. The `mergeURL()` function uses `path.Join()` to combine paths
2. `path.Join()` calls `path.Clean()`, which removes trailing slashes
3. Even when explicitly specified in YAML, trailing slashes are removed for APIs that require them
4. No workaround exists (currently impossible to test such APIs with runn)

### Proposed Solution

Add a `preserveTrailingSlash` option at the step level:

```yaml
runners:
  api: https://api.example.com
steps:
  - api:
      /api/users/:
        get:
          preserveTrailingSlash: true  # Add this option
```

When this option is specified, the trailing slash will be preserved.

#### Why Use an Option?

1. **Backward Compatibility**: Existing runbooks remain unaffected
2. **Minimal Change**: Cases requiring trailing slashes are limited
3. **Explicit Intent**: Opt-in approach - enable only when needed

#### Why Step-Level?

[...]

PR #1326, opened 2025-10-14
 fix: trailing slash removal in HTTP request paths

helloworld753315:
## Summary

`path.Join()` removes trailing slashes, but user's explicitly specified trailing slashes should be preserved to support APIs that require them.

ref: #1323

## Changes

- Modified `mergeURL()` to preserve trailing slashes when explicitly specified
- Updated test cases

## Test

Verified by `go test -run TestMergeURL -v` - all tests pass.


k1LoW (approved):
GREAT!! Thank you!!

PR #1450, opened 2026-04-11
Cookies are not propagated between parent and included runbooks (include scenario) 

### LINKED ISSUE #1449: Cookies are not propagated between parent and included runbooks (include scenario)
## Overview

When running an included runbook (`include:` scenario), cookies set via `Set-Cookie` response headers in the included runbook are **not** propagated back to the parent runbook, and vice versa. This makes it difficult or impossible to maintain session continuity across steps that span included (child) and parent (main) runbooks, even with `useCookie: true` set for runners in both contexts.

---

## Current Behavior (as of commit c6e4094)
- Each runbook (parent and include) instantiates its own `operator` and `store`.
- Cookies are managed in `store.cookies` ([internal/store/store.go](https://github.com/k1LoW/runn/blob/c6e40943935f23a117521530609aaff275d64161/internal/store/store.go#L261-L281)), and are not copied or merged between parent and child.
- See [include.go:newNestedOperator](https://github.com/k1LoW/runn/blob/c6e40943935f23a117521530609aaff275d64161/include.go#L219), and [includeRunner.run](https://github.com/k1LoW/runn/blob/c6e40943935f23a117521530609aaff275d64161/include.go#L184).
- The `RecordCookie` function always overrides any existing cookies (does not merge, just replaces) ([internal/store/store.go](https://github.com/k1LoW/runn/blob/c6e40943935f23a117521530609aaff275d64161/internal/store/store.go#L261)).
- As a result:
    - Parent cookies are not visible to the included runbook.
    - Cookies set in the included runbook do not reach the parent context.
    - The most recent Set-Cookie response always replaces all cookies, so cookies are not accumulated like a browser/session would do.

---

## Expected (Browser-Like) Behavior
- Cookies set during HTTP(S) requests should accumulate and persist throughout the entire runbook execution chain (including across included runbooks), just like a browser cookie jar.
- When using `useCookie: true`, any cookies set (unless expired) should be available to both parent and any subsequently included/including runbooks.

---

## Technical Analysis
- **Root Cause:** Cookies are stored only inside each `store` instance, not shared/copied between operators.
    - There is no explicit propagation or merging:
        - At [include.go:newNestedOperator](https://github.com/k1LoW/runn/blob/c6e40943935f23a117521530609aaff275d64161/include.go#L219), only variables, kv, etc. are shared, not cookies.
        - At the end of include execution ([includeRunner.run](https://github.com/k1LoW/runn/blob/c6e40943935f23a117521530609aaff275d64161/include.go#L184)), cookies are not merged back to the parent operator.
    - The central cookie management logic ([internal/store/store.go:RecordCookie](https://github.com/k1LoW/runn/blob/c6e40943935f23a117521530609aaff275d64161/internal/store/store.go#L261-L281)) replaces cookies map, instead of merging existing ones akin to browser behavior.
- This design results in lack of session propagation, and also causes loss of cookies when multiple domains or step-wise cookie accumulation are expected.

---

## Proposal (for feedback)

I would like to confirm if the following direction aligns with the intended design, or if there are any edge cases or design considerations that would warn against it:

**1. Migrate cookie management to a browser-like (accumulative, merging) model:**
- Change `RecordCookie` to merge/accumulate cookies by domain/key, removing only expired ones, similar to browser cookie engines.
- Update tests (e.g. `TestRecordCookie`) to match this behavior.

**2. Propagate cookies parent→child and merge back child→parent for include:**
- On `newNestedOperator`, deep copy the parent store's cookies into the child.
- After `includeRunner.run`, merge the child's cookies back into the parent.
- This allows seamless session/cookie usage regardless of step belonging to parent or child/included runbook (mirrors actual browser session behavior).

---

## References to code
- [internal/store/store.go#L261-L281 (RecordCookie)](https://github.com/k1LoW/runn/blob/c6e40943935f23a117521530609aaff275d64161/internal/store/store.go#L261-L281)
- [include.go#L219 (newNestedOperator)](https://github.com/k1LoW/runn/blob/c6e40943935f23a117521530609aaff275d64161/include.go#L219)
- [include.go#L184 (includeRunner.run)](https://github.com/k1LoW/runn/blob/c6e40943935f23a117521530609aaff275d64161/include.go#L184)

## Risk/Impact
- If any scenario relies on the _current_ "overwrite cookies per Set-Cookie" behavior, those tests/scenarios will need to be updated, but this is generally not standard HTTP client behavior and may be considered a bug.
- Unintended cookie propagation between otherwise-isolated runbooks may need additional opt-out mechanisms if runbook scoping is important.

k2tzumi:
close #1449

### DIFF
--- a/include.go
+++ b/include.go
@@ -18,10 +18,10 @@ type includeRunner struct {
 }
 
 type includeConfig struct {
-	path     string
-	vars     map[string]any
-	skipTest bool
-	force    bool
+	path       string
+	vars       map[string]any
+	skipTest   bool
+	force      bool
 	step       *step
 	inline     bool
 	desc       string
@@ -211,6 +211,7 @@ func (rnr *includeRunner) run(ctx context.Context, oo *operator, s *step) error 
 		}
 		rnr.runResults = append(rnr.runResults, ooo.runResult)
 	}
+	o.store.MergeCookies(oo.store.Cookies())
 	o.record(s.idx, oo.store.ToMapForIncludeRunner())
 	return nil
 }
--- a/include.go
+++ b/include.go
@@ -210,6 +210,7 @@ func (rnr *includeRunner) run(ctx context.Context, oo *operator, s *step) error 
 			oo.store.MergeCookies(ooo.store.Cookies())
 			return newIncludedRunErr(err)
 		}
+		oo.store.MergeCookies(ooo.store.Cookies())
 		rnr.runResults = append(rnr.runResults, ooo.runResult)
 	}
 	o.store.MergeCookies(oo.store.Cookies())
--- a/include.go
+++ b/include.go
@@ -207,6 +207,7 @@ func (rnr *includeRunner) run(ctx context.Context, oo *operator, s *step) error 
 		ooo.parent = oo.parent
 		if err := ooo.run(ctx); err != nil {
 			rnr.runResults = append(rnr.runResults, ooo.runResult)
+			oo.store.MergeCookies(ooo.store.Cookies())
 			return newIncludedRunErr(err)
 		}
 		rnr.runResults = append(rnr.runResults, ooo.runResult)
--- a/include.go
+++ b/include.go
@@ -234,6 +234,7 @@ func (o *operator) newNestedOperator(parent *step, opts ...Option) (*operator, e
 	oo.capturers = o.capturers
 	oo.parent = parent
 	oo.store.SetParentVars(o.store.ToMap())
+	oo.store.MergeCookies(o.store.Cookies())
 	oo.store.SetKV(o.store.KV())
 	oo.store.SetRunNIndex(o.store.RunNIndex())
 	oo.store.SetMaskRule(o.store.MaskRule())
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -156,7 +156,12 @@ func (s *Store) MergeCookies(cookies map[string]map[string]*http.Cookie) {
 			s.cookies[domain] = map[string]*http.Cookie{}
 		}
 		for name, cookie := range domainCookies {
-			s.cookies[domain][name] = cloneCookie(cookie)
+			if cookie != nil && !cookie.Expires.IsZero() && cookie.Expires.Before(time.Now()) {
+				// Remove expired cookie
+				delete(s.cookies[domain], name)
+			} else {
+				s.cookies[domain][name] = cloneCookie(cookie)
+			}
 		}
 	}
 }
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -116,6 +116,51 @@ func (s *Store) Cookies() map[string]map[string]*http.Cookie {
 	return s.cookies
 }
 
+func cloneCookie(c *http.Cookie) *http.Cookie {
+	if c == nil {
+		return nil
+	}
+	cc := *c
+	return &cc
+}
+
+func cloneCookies(cookies map[string]map[string]*http.Cookie) map[string]map[string]*http.Cookie {
+	if cookies == nil {
+		return nil
+	}
+	cloned := make(map[string]map[string]*http.Cookie, len(cookies))
+	for domain, domainCookies := range cookies {
+		if domainCookies == nil {
+			cloned[domain] = map[string]*http.Cookie{}
+			continue
+		}
+		cm := make(map[string]*http.Cookie, len(domainCookies))
+		for name, cookie := range domainCookies {
+			cm[name] = cloneCookie(cookie)
+		}
+		cloned[domain] = cm
+	}
+	return cloned
+}
+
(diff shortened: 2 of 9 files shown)


include.go
```diff
@@ -211,6 +211,7 @@ func (rnr *includeRunner) run(ctx context.Context, oo *operator, s *step) error
 		}
 		rnr.runResults = append(rnr.runResults, ooo.runResult)
 	}
```
k1LoW, line 215: When the child runbook fails, this line is not reached because of the early return at L210. If the parent uses `force: true`, it will continue execution but without the cookies that the child had already set. Should `MergeCookies` be called before the early return as well, or use `defer`?

internal/store/store.go
```diff
@@ -150,6 +195,9 @@ func (s *Store) RecordBindVar(k string, v any, sm map[string]any) error {
 
 func (s *Store) SetParentVars(vars map[string]any) {
 	s.parentVars = vars
+	if cookies, ok := vars[RootKeyCookie].(map[string]map[string]*http.Cookie); ok {
+		s.MergeCookies(cloneCookies(cookies))
+	}
```
k1LoW: Cookie propagation as a side effect of `SetParentVars` is not obvious from the method name. Consider moving this logic to `newNestedOperator` in `include.go` as an explicit call:

```go
oo.store.MergeCookies(o.store.Cookies())
```

This makes the cookie propagation visible at the orchestration layer rather than hidden inside a setter.

internal/store/store.go
```diff
@@ -150,6 +195,9 @@ func (s *Store) RecordBindVar(k string, v any, sm map[string]any) error {
 
 func (s *Store) SetParentVars(vars map[string]any) {
 	s.parentVars = vars
+	if cookies, ok := vars[RootKeyCookie].(map[string]map[string]*http.Cookie); ok {
+		s.MergeCookies(cloneCookies(cookies))
```
k1LoW: `MergeCookies` already calls `cloneCookie` for each cookie internally. Wrapping the argument with `cloneCookies` here results in a double deep copy. You can simplify this to:

```go
s.MergeCookies(cookies)
```

k1LoW:
Pre-existing issue (not introduced by this PR): `setCookieHeader` is not called in the `rnr.handler` (httptest) path at `http.go:533`, so cookies are not sent when using `http.Handler`-based runners. This may affect test scenarios that rely on cookie propagation with `httptest`.

k1LoW:
The overall direction of this PR looks good to me! I've added (and am still working on) AI-assisted reviews, but I'm fine with making final adjustments after merging.

include.go
k2tzumi: Thanks for the great catch. I updated the include error path so cookies are merged before returning on child failure. This preserves child-set cookies even when the parent continues with force: true.

internal/store/store.go
k2tzumi: Agreed. I removed cookie propagation from SetParentVars and made it explicit in newNestedOperator. This makes the flow visible at the orchestration layer and avoids hidden setter side effects.

internal/store/store.go
k2tzumi: Good point, thanks. The redundant copy was removed as part of moving propagation out of SetParentVars. Also, the now-unused clone helper was deleted.

k2tzumi:
https://github.com/k1LoW/runn/pull/1450#issuecomment-4230561123

Thanks for highlighting this. I confirmed it is pre-existing and out of scope for this PR. I will handle it separately in a follow-up issue/PR so this change stays focused.

include.go
```diff
@@ -207,10 +207,12 @@ func (rnr *includeRunner) run(ctx context.Context, oo *operator, s *step) error
 		ooo.parent = oo.parent
 		if err := ooo.run(ctx); err != nil {
 			rnr.runResults = append(rnr.runResults, ooo.runResult)
+			oo.store.MergeCookies(ooo.store.Cookies())
 			return newIncludedRunErr(err)
 		}
 		rnr.runResults = append(rnr.runResults, ooo.runResult)
```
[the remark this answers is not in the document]
k2tzumi: I intentionally kept this as a one-line change in includeRunner.run to minimize behavioral impact while fixing the reported propagation gap.
 
Reasoning:
- The issue in this thread is that cookies from a successfully executed child operator were not merged into the include-level store before the next operator context.
- Adding a merge right after successful execution addresses that gap directly with the smallest possible change.
- The existing error-path merge and final parent merge already cover failure and end-of-include propagation, so this addition makes success-path behavior symmetric without restructuring the flow.
- This keeps the fix easy to review, low risk for regressions, and scoped to the specific concern raised in this discussion.

If we want stronger chain-wide guarantees across more complex include and needs topologies, we can discuss an additional follow-up refactor separately.

k1LoW (approved):
LGTM!

PR #1514, opened 2026-08-29
Add --debug-on-failure option

udzura:
## Summary

Add `--debug-on-failure` to print runner diagnostics only for failed steps while preserving the normal output for successful steps.

The same behavior is also available to library users through `runn.DebugOnFailure`.

## Motivation

`--verbose` reports failed steps and assertions, but does not include captured runner output such as Exec stdout and stderr.

`--debug` includes those diagnostics, but prints them for every step, which can make successful CI output unnecessarily noisy.

`--debug-on-failure` provides the diagnostics needed to investigate failures without changing the normal successful output.

## Behavior

- Writes failure diagnostics to stderr.
- Includes the command, stdout, and stderr for Exec and SSH runners.
- Includes stdin for Exec runners and the captured request/response details for other runners.
- Always masks values configured through `secrets:`.
- Limits buffered diagnostics to 1 MiB per step and reports truncation.
  - This is an initial design choice, and the limit and truncation strategy are open to discussion.
- Shows only the last iteration of a failed loop step.
- Shows only the actual failed inner step for included runbooks.
- Serializes diagnostic blocks from concurrently running runbooks to avoid mixed output.
- Gives `--debug` precedence when both options are specified.
- Leaves `liveOutput: true` behavior unchanged.

Exec runners do not treat a non-zero exit code as a failure by itself. A test such as `test: current.exit_code == 0` is still required when the exit code should fail the step.

## Example

```console
runn run path/to/*/*.yml --verbose --debug-on-failure
```

Successful steps keep the normal output. When a step fails, diagnostics are written in a distinct block:

```
-----START FAILURE DIAGNOSTICS-----
Runbook: runbooks/verify.yml
Step: 0
Runner: exec (exec)

-----START COMMAND-----
ruby verify.rb basic
-----END COMMAND-----
-----START STDOUT-----
ok 1 - GET route
ok 2 - named route parameter
not ok 3 - POST form parameters
ok 4 - response status and header
ok 5 - not found
1..5

-----END STDOUT-----
-----START STDERR-----
  AssertionFailure: body: expected "420", got "42"
/Users/udzura/...

-----END STDERR-----
-----END FAILURE DIAGNOSTICS-----
```

## Testing

```
go test . -run 'TestDebugOnFailure|TestDebugTakesPrecedence' -count=1
go test -race . -run 'TestDebugOnFailure|TestDebugTakesPrecedence' -count=1
go vet ./...
```

Also, 

- Confirmed successful runs produce normal verbose output with no stderr diagnostics.
- Confirmed failed Exec steps produce only the relevant command, stdout, and stderr diagnostics on stderr.

### DIFF
--- a/failure_debugger.go
+++ b/failure_debugger.go
@@ -1,0 +1,129 @@
+package runn
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"sync"
+
+	"github.com/k1LoW/maskedio"
+)
+
+const failureDebugMaxBytes = 1 << 20
+
+type failureDebugCoordinator struct {
+	mu sync.Mutex
+}
+
+func newFailureDebugCoordinator() *failureDebugCoordinator {
+	return &failureDebugCoordinator{}
+}
+
+func (c *failureDebugCoordinator) write(out *maskedio.Writer, b []byte) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	_, err := out.Write(b)
+	return errors.Join(err, out.Flush())
+}
+
+type limitedBuffer struct {
+	buf       bytes.Buffer
+	max       int
+	truncated bool
+}
+
+func newLimitedBuffer(maxBytes int) *limitedBuffer {
+	return &limitedBuffer{max: maxBytes}
+}
+
+func (b *limitedBuffer) Write(p []byte) (int, error) {
+	n := len(p)
+	remaining := b.max - b.buf.Len()
+	if remaining <= 0 {
+		b.truncated = b.truncated || n > 0
+		return n, nil
+	}
+	if len(p) > remaining {
+		_, _ = b.buf.Write(p[:remaining])
+		b.truncated = true
+		return n, nil
+	}
+	_, _ = b.buf.Write(p)
+	return n, nil
+}
+
+func (b *limitedBuffer) Bytes() []byte {
+	return b.buf.Bytes()
+}
+
+func (b *limitedBuffer) Len() int {
+	return b.buf.Len()
+}
+
+func (b *limitedBuffer) Reset() {
+	b.buf.Reset()
+	b.truncated = false
+}
+
+type failureDebugger struct {
+	*debugger
+	buf         *limitedBuffer
+	out         *maskedio.Writer
+	coordinator *failureDebugCoordinator
+}
+
+func newFailureDebugger(out *maskedio.Writer, coordinator *failureDebugCoordinator) *failureDebugger {
+	buf := newLimitedBuffer(failureDebugMaxBytes)
+	return &failureDebugger{
+		debugger:    NewDebugger(buf),
+		buf:         buf,
+		out:         out,
+		coordinator: coordinator,
+	}
+}
+
+func (d *failureDebugger) CaptureResultByStep(_ Trails, result *RunResult) {
+	defer d.buf.Reset()
+
+	sr := latestStepResult(result.StepResults)
+	if sr == nil || sr.Err == nil || d.buf.Len() == 0 {
+		return
+	}
+
+	var out bytes.Buffer
+	_, _ = fmt.Fprintln(&out, "-----START FAILURE DIAGNOSTICS-----")
+	_, _ = fmt.Fprintf(&out, "Runbook: %s\n", normalizePath(result.Path))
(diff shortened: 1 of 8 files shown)


failure_debugger.go
```diff
@@ -0,0 +1,129 @@
...
+	n := len(p)
+	remaining := b.max - b.buf.Len()
+	if remaining <= 0 {
+		b.truncated = b.truncated || n > 0
+		return n, nil
+	}
+	if len(p) > remaining {
+		_, _ = b.buf.Write(p[:remaining])
+		b.truncated = true
+		return n, nil
```
k1LoW: Truncation here cuts the byte stream at an arbitrary offset, but masking is only applied later, when the assembled block is written to `d.out` (`*maskedio.Writer`) in `CaptureResultByStep`. Since `maskedio` matches whole literals, a `secrets:` value that straddles the 1 MiB boundary is emitted as an unmasked prefix. The same cut can also split a UTF-8 rune.

Backing off to the last newline before the cut would close this for single-line secrets: a value crossing the boundary then falls entirely inside the discarded partial line, and the retained bytes stay valid UTF-8.

```go
if len(p) > remaining {
	w := p[:remaining]
	if i := bytes.LastIndexByte(w, '\n'); i >= 0 {
		w = w[:i+1]
	}
	_, _ = b.buf.Write(w)
	b.truncated = true
	return n, nil
}
```

Masking at buffer-write time instead is the other option, but it needs care. `secrets:` keywords are registered lazily during the run (`internal/store/store.go:463`), so a writer built at `New()` time would miss values that only become known mid-run. `maskedio.Writer.NewSameWriter` copies the keyword list and would therefore be the wrong choice; `Rule.NewWriter` shares the `*Rule`. It would also need `DisableAutoFlush()` plus an explicit `Flush()`, since `Write` otherwise spawns a goroutine that flushes into the buffer ~100µs later and races the read here.

k1LoW: Truncation here cuts the byte stream at an arbitrary offset, but masking is only applied later, when the assembled block is written to `d.out` (`*maskedio.Writer`) in `CaptureResultByStep`. `maskedio` substitutes literal keywords through a `strings.Replacer`, so a `secrets:` value that straddles the 1 MiB boundary is left behind as an unmasked prefix. The same cut can also split a UTF-8 rune.

Backing off to the last newline is the cheap fix, but it has to be applied to the buffer rather than to the incoming chunk, and the buffer has to be sealed afterwards. Trimming only `p[:remaining]` still leaks whenever the first half of the value arrived in an earlier `Write`: `"ab\nSUPER"` followed by `"SECRET\nxx"` leaves `"ab\nSUPERSE"` in the buffer. And trimming without sealing re-opens room under the cap, so the next chunk appends a fresh partial line.

```go
func (b *limitedBuffer) Write(p []byte) (int, error) {
	n := len(p)
	if b.truncated {
		return n, nil
	}
	remaining := b.max - b.buf.Len()
	if len(p) <= remaining {
		_, _ = b.buf.Write(p)
		return n, nil
	}
	if remaining > 0 {
		_, _ = b.buf.Write(p[:remaining])
	}
	b.truncated = true
	// WHY: a secrets: value straddling the cut would survive as a prefix that the
	// literal replacement in maskedio can no longer match, so drop the partial line.
	if i := bytes.LastIndexByte(b.buf.Bytes(), '\n'); i >= 0 {
		b.buf.Truncate(i + 1)
	} else {
		b.buf.Reset()
	}
	return n, nil
}
```

The trade-off is the no-newline case: diagnostics that are a single line longer than the cap now come out empty instead of truncated. A JSON response body over 1 MiB on one line is not unusual, so it is worth deciding that explicitly rather than by accident.

Masking into the buffer instead would sidestep the trade-off, but it needs care. `secrets:` keywords keep being registered while the run proceeds (`internal/store/store.go:447-463`, reached from `ToMap*`), so `maskedio.Writer.NewSameWriter` is the wrong tool here because it copies the keyword slice at construction time; `Rule.NewWriter` shares the `*Rule`. It would also want `DisableAutoFlush()` plus an explicit `Flush()` before `d.buf.Bytes()` is read, because `Write` defers a flush by roughly 100µs whenever the tail of a chunk looks like the beginning of a keyword.

k1LoW (approved):
@udzura Great work!!! Thank you!! 
I plan to release it after making a few adjustments. Please give me a moment.

PR #1525, opened 2026-09-17
Cache process environment in Store

go-kazuhiko-yamashita:
## What

`Store.ToMap()` rebuilds the `env` map from `os.Environ()` on every call.
`ToMap()` runs for every expression evaluation, so a runbook with many steps
pays this allocation over and over. This caches the map on the `Store`.

```go
func (s *Store) envMap() map[string]string {
	if s.env == nil {
		s.env = envMap()
	}
	return s.env
}
```

`ToMap` / `ToMapForIncludeRunner` / `ToMapForDbg` now use it.

## Why this is safe

- The cache is per-`Store`, not process-global. A newly created `Store` reads
  the current environment, so existing tests using `t.Setenv` keep working.
- `os.Setenv` is only called from `LoadEnvFile`, which runs before the run
  starts. The environment does not change during a run.
- `Store` has no mutex (single-goroutine by design), so this introduces no
  new race.

## Assumption

Callers must not mutate the returned `store["env"]` map.

## Test

`go test ./internal/store/...` — 45 passed.


k1LoW:
@pyama86 Thank you!!!!!

k1LoW:
I agree with your sense of the challenge, and the caching strategy itself looks good.

Since we also need to consider consistency with things like runn.LoadEnvFile, I plan to make adjustments after merging this PR.

Thank you!

PR #1527, opened 2026-09-17
docs: add design doc for handling of the process environment

k1LoW:
## Summary

- Add `docs/designs/env.md`, which writes down how runn handles the process environment: `env` is a snapshot taken at the start of each run and immutable for its duration, included runbooks inherit the parent's snapshot, the `runn` package never calls `os.Setenv`, and `--env-file` is the one place that does, once, at CLI startup
- The document also records why an operator-scoped `runn.EnvFile()` was rejected: it could reach only `env.*`, `${}` expansion and the exec runner, not the `RUNN_*` settings or dependencies that read the environment directly
- Link it from the `env` row of the variables table in the README

Refs #1526

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


k1LoW:
@k2tzumi Please review this when you have a moment.

